PR #4, opened 2018-08-22
copy initial proposed API and readme from sandbox

### LINKED ISSUE #3: Copy over the sandbox APIs and the readme file
Copy the API and readme from the MicroProfile sandbox location,
https://github.com/eclipse/microprofile-sandbox/tree/master/proposals/concurrency/

For the pom files, I am copying from existing MicroProfile projects so that we get the same level of validation.

The readme file will be copied identically, except with the addition of a license (required by the validation) and the addition of the gitter link.

The API files will be copied identically except for the removal of unused imports, which is required by the validation.

njr-11:
pull fixes #3

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.concurrent;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * <p>Configures an injected <code>ManagedExecutor</code> instance.</p>
+ *
+ * <p>Example usage:</p>
+ * <pre><code> &commat;Inject &commat;ManagedExecutorConfig(context=ThreadContext.CDI, maxAsync=5)
+ * ManagedExecutor executor;
+ * ...
+ * </code></pre>
+ */
+@Retention(RUNTIME)
+@Target(FIELD)
+public @interface ManagedExecutorConfig {
+    /**
+     * <p>Defines the set of thread context types to capture from the thread
+     * that creates a dependent stage (or that submits a task) and which to
+     * propagate to the thread where the action or task executes.</p>
+     *
+     * <p>The default set of thread context types is those required by the
+     * EE Concurrency spec, plus CDI.</p>
+     *
+     * <p>Constants for specifying some of the core context types are provided
+     * on {@link ThreadContext}. Other thread context types must be defined
+     * by the specification that defines the context type or by a related
+     * MicroProfile specification.</p>
+     *
+     * <p>Inclusion of a thread context type with prerequisites implies
+     * inclusion of the prerequisites, even if not explicitly specified.</p>
+     *
+     * <p>Thread context types which are not otherwise included in this set
+     * are removed from the thread of execution for the duration of the
+     * action or task.</p>
+     */
+    String[] context() default { ThreadContext.APPLICATION, ThreadContext.CDI, ThreadContext.SECURITY };
+
+    /**
+     * <p>Establishes an upper bound on the number of async completion stage
+     * actions and async executor tasks that can be running at any given point
+     * in time. Async actions and tasks remain queued until capacity is available
+     * to execute them.</p>
+     *
+     * <p>The default value of <code>-1</code> indicates no upper bound,
+     * although practically, resource constraints of the system will apply.</p>
+     */
+    int maxAsync() default -1;
+
+    /**
+     * <p>Establishes an upper bound on the number of async actions and async tasks
+     * that can be queued up for execution. Async actions and tasks are rejected
+     * if no space in the queue is available to accept them.</p>
+     *
+     * <p>The default value of <code>-1</code> indicates no upper bound,
+     * although practically, resource constraints of the system will apply.</p>
+     */
+    int maxQueued() default -1;
+}
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ActiveThreadContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ActiveThreadContext.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.concurrent.spi;
+
+/**
+ * <p>Represents context that is applied to a particular thread, along with any
+ * state that is associated with it or that is necessary for restoring the
+ * previous context afterward.</p>
+ *
+ * <p>When the context is no longer needed on the thread, the
+ * <code>ManagedExecutor</code> or <code>ThreadContext</code> must invoke the
+ * <code>end</code> method.</p>
+ */
+public interface ActiveThreadContext {
+    /**
+     * <p>Invoked by the <code>ManagedExecutor</code> or
+     * <code>ThreadContext</code> to remove the thread context represented by
+     * this <code>ActiveThreadContext</code> instance and restore the previous
+     * context that was on the thread before the <code>ActiveThreadContext</code>
+     * was started on the thread. The <code>ManagedExecutor</code> or
+     * <code>ThreadContext</code> must invoke the <code>end</code> method exactly
+     * once for each <code>ActiveThreadContext</code> instance that it creates.</p>
+     *
+     * <p>Typically, patterns such as the following will be observed:</p>
+     * <pre><code>
+     * activeContextA1 = contextSnapshotA.begin();
+     * activeContextB1 = contextSnapshotB.begin();
+     * activeContextC1 = contextSnapshotC.begin();
+     * ...
+     * activeContextC1.end();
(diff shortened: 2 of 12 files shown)


FroMage:
LGTM: we can start with that and work through remaining issues afterwards.

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
```diff
@@ -0,0 +1,81 @@
...
+     * MicroProfile specification.</p>
+     *
+     * <p>Inclusion of a thread context type with prerequisites implies
+     * inclusion of the prerequisites, even if not explicitly specified.</p>
+     *
+     * <p>Thread context types which are not otherwise included in this set
+     * are removed from the thread of execution for the duration of the
+     * action or task.</p>
+     */
+    String[] context() default { ThreadContext.APPLICATION, ThreadContext.CDI, ThreadContext.SECURITY };
```
Emily-Jiang: Should ALL be the default?

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ActiveThreadContext.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * <p>Represents context that is applied to a particular thread, along with any
+ * state that is associated with it or that is necessary for restoring the
+ * previous context afterward.</p>
+ *
+ * <p>When the context is no longer needed on the thread, the
+ * <code>ManagedExecutor</code> or <code>ThreadContext</code> must invoke the
+ * <code>end</code> method.</p>
+ */
+public interface ActiveThreadContext {
```
Emily-Jiang: I would say this is a controller. How about ThreadContextController?

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ActiveThreadContext.java
```diff
@@ -0,0 +1,73 @@
...
+     * ...
+     * activeContextC2.end();
+     * activeContextB2.end();
+     * ...
+     * activeContextB1.end();
+     * activeContextA1.end();
+     * </code></pre>
+     *
+     * @throws IllegalStateException if invoked more than once on the same instance.
+     */
```
Emily-Jiang: There should a method e.g. getState() to return whether it is active or not. end() can check whether it is active or not. If it is, end it. If not, ignore it.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextProvider.java
```diff
@@ -0,0 +1,143 @@
...
+     * @param props provided for compatibility with EE Concurrency spec, which allows
+     *        for specifying a set of 'execution properties' when capturing thread context.
+     *        Thread context providers that don't supply or use execution properties
+     *        can ignore this parameter.
+     * @return immutable snapshot of the provided type of context, captured from the
+     *         current thread. NULL must be returned if the thread context provider
+     *         does not support capturing context from the current thread and
+     *         propagating it to other threads.
+     */
+    ThreadContextSnapshot currentContext(Map<String, String> props);
```
Emily-Jiang: Should this be String, String? Maybe String, Object has more coverage.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextSnapshot.java
```diff
@@ -0,0 +1,54 @@
...
+     * previous context is restored on the thread. If the invoker sequentially
+     * begins multiple <code>ActiveThreadContext</code> instances on a thread,
+     * it must invoke the corresponding <code>end</code> methods in reverse
+     * order.</p>
+     *
+     * @return immutable snapshot of the provided type of context, typically
+     *         either captured from a thread or otherwise representing an
+     *         empty/default context.
+     */
+    ActiveThreadContext begin();
```
Emily-Jiang: Can you put this method to ThreadContextController?

Emily-Jiang:
@njr-11 please take a look at my comments.

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
njr-11: The prior default was chosen to match the EE Concurrency requirement for only a specific set of context types to be propagated by default, where
 @Inject @ManagedExecutorConfig ManagedExecutor executor1;
would create an instance with the same context propagation as
 @Inject ManagedExecutor executor2; 
which would also have the same context propagation as EE Concurrency's default instance,
 @Resource ManagedExecutorService executor3;
The idea was to allow a container that is both MP Concurrency and EE Concurrency to return the same instance from all of these.

That said, it would be possible to instead default to ALL and require the use of different instance than the EE Concurrency default instance if that would be more useful.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextProvider.java
njr-11: It is `Map<String, String>` for compatibility with EE Concurrency execution properties. Switching to Object would force code to deal with unwanted casting.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextSnapshot.java
njr-11: Leaving it here enforces the contract that each ActiveThreadContext/ThreadContextController represents exactly one contextualized execution.  If we move it, then the implementer must deal with the possibility that the user tries to invoke begin multiple times on the same instance, or tries to re-begin one that has already ended.  I can't see a reason to place this additional burden/lack of clarity upon the provider of thread context without any clear benefit that would be achieved by doing so.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ActiveThreadContext.java
njr-11: Sure, we can add this, maybe as a boolean isActive() given that there are only 2 states.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ActiveThreadContext.java
njr-11: I like how the name ThreadContextController starts with ThreadContext in the same way that ThreadContextProvider and ThreadContextSnapshot do, which will cause them to naturally be listed together when interfaces are shown alphabetically.  ActiveThreadContext seems more descriptive/clear, except could be confusing after it is ended, so I'm willing to change it as you suggest.

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
njr-11: Thinking about this some more, defaulting to ThreadContext.ALL would mean that if someone implemented a provider for transaction context that was capable of propagating transactions, then those would also propagate by default.  I don't think we want to go that far.  It seems better for performance and predictability that we default to the more limited set of core contexts.  The user can always specify ALL if they want it.

njr-11:
@Emily-Jiang thanks for the review.  I have added a pull with several code review fixes (per comments above) as well as a couple of spelling and punctuation fixes that Nathan M and myself spotted while proofreading.

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
Emily-Jiang: ```Thinking about this some more, defaulting to ThreadContext.ALL would mean that if someone implemented a provider for transaction context that was capable of propagating transactions, then those would also propagate by default.```

If an user provide a transaction context, surely the default one will be overriden. You won't need two of them.

njr-11: The 'someone' that I was referring wasn't intended to mean the end user.  I was referring to a container implementation of transactions (the transaction manager).  In the event that an end user did provide a conflicting implementation of one of already-provided context types, I think we discussed that would be considered an error path for now.  Allowing an override of context provided by someone else seems outright dangerous, and in the case of transactions would probably lead to data integrity issues.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextProvider.java
```diff
@@ -0,0 +1,143 @@
...
+ * <code>ThreadContext</code> interfaces.</p>
+ *
+ * <p>The <code>ThreadContextProvider</code> implementation and related classes are
+ * packaged within the third party provider's JAR file. The implementation is made
+ * discoverable via the standard <code>ServiceLoader</code> mechanism. The JAR file
+ * that packages it must include a file of the following name and location,</p>
+ *
+ * <code>META-INF/services/org.eclipse.microprofile.concurrent.spi.ThreadContextProvider</code>
+ *
+ * <p>The content of the aforementioned file must be a single line that specifies the
```
FroMage: Actually it may be more than one line, each with a different implementation of the interface.

FroMage:
Can we please merge these two PRs ASAP and fix whatever remaining issue afterwards? It really doesn't help implementation to keep these in PRs.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextProvider.java
```diff
@@ -39,8 +39,9 @@
  *
  * <code>META-INF/services/org.eclipse.microprofile.concurrent.spi.ThreadContextProvider</code>
  *
- * <p>The content of the aforementioned file must be a single line that specifies the
- * fully qualified name of the <code>ThreadContextProvider</code> implementation.</p>
+ * <p>The content of the aforementioned file must be one or more lines, each specifying
+ * the fully qualified name of a <code>ThreadContextProvider</code> implementation
+ * that is provided within the JAR file.</p>
```
FroMage: Thanks!

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextProvider.java
njr-11: Thanks - good catch.  We definitely want to allow for multiple.  I have added a commit correcting the language here to indicate at least one instead of single.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextController.java
```diff
@@ -0,0 +1,82 @@
...
+    void endContext() throws IllegalStateException;
+
+    /**
+     * Indicates whether or not the context managed by this controller has been
+     * ended on this thread by invoking <code>endContext</code>.
+     *
+     * @return true if <code>endContext</code> was invoked on this controller,
+     *         otherwise false.
+     */
+    boolean isEnded();
```
FroMage: Note that this prevents implementing this interface as a `@FunctionalInterface` as I had before:

```
	public ThreadContextSnapshot currentContext(Map<String, String> props) {
		MyContext capturedContext = MyContext.get();
		return () -> {
			MyContext movedContext = MyContext.get();
			MyContext.set(capturedContext);
			return () -> {
				MyContext.set(movedContext);
			};
		};
	}

```

njr-11: Thank you for spotting this as well.  I was previously willing to accommodate the request to have an isEnded method when I saw no benefit either way for including or not including it.  However, as you have pointed out, this makes it no longer be a functional interface.  I have added a commit to remove the isEnded method.

FroMage: Yeah, I noticed it by accident, but I thought it made a good argument in favour of not having that extra method. I suppose we should also add `@FunctionalInterface` to the interfaces (at least if/until we decide to add methods to it) to make sure that we don't add any new method without just cause.

njr-11: Good idea. `@FunctionalInterface` is added to both ThreadContextController and ThreadContextSnapshot.  I think as long as they are able to remain functional interfaces for the initial release, they should hopefully remain that way.  We won't want to cause a lot of churn in the SPI once implementations of it start to be provided.

PR #8, opened 2018-09-06
possible builder pattern to address issue #6

### LINKED ISSUE #6: concurrency: support `ThreadContext` lookup without CDI
In order for mp-concurrency to replace [Reactive Contexts](https://reactiverse.io/reactive-contexts/) which is currently used by RESTEasy to provide context propagation, I suspect we will need to be able to obtain a `TreadContext` implementation without CDI, because RESTEasy-core does not depend on CDI.

Keeping the dependencies of mp-concurrency would allow it to be more broadly used. A CDI implementation could be provided as an optional dependency.

Copied from https://github.com/eclipse/microprofile-sandbox/issues/18

njr-11:
In this pull, which fixes #6, I have copied the ManagedExecutorConfig/ThreadContextConfig annotations used for CDI into ManagedExecutorBuilder/ThreadContextBuilder interfaces that can be used independently of CDI.  Will need to go over with the spec group if this is an approach we want to pursue.


FroMage:
LGTM: we can start with that and work through remaining issues afterwards.

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+     *         and the {@link #unchanged} set.
+     */
+    ThreadContext build();
+
+    /**
+     * Creates a new <code>ThreadContextBuilder</code> instance.
+     *
+     * @return a new <code>ThreadContextBuilder</code> instance.
+     */
+    public static ThreadContextBuilder create() {
```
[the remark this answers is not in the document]
njr-11: I've updated it to rename the create method to instance on both builders.  However, please hold off on merging this pull because it still needs discussion.  I don't know if this will actually solve the issues raised by Stephane and I'm thinking about a different solution that might be better, in which case we wouldn't necessarily require these builders at all.

njr-11:
I have attached a commit, https://github.com/eclipse/microprofile-concurrency/pull/8/commits/08e0f07dda2b41024596e864fb94c612901b5b8e, which addresses the review comments to #12 that pertain to this pull.  Please review and let us know if it is now acceptable to merge.

njr-11:
@FroMage could you review the commit referenced above with the fix to the naming/doc inconsistencies that you identified and let us know if it is good to merge. Thanks!

FroMage:
Yes, this is better indeed.

genie-microprofile:

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+     * inclusion of the prerequisites, even if not explicitly specified.</p>
+     *
+     * <p>Thread context types which are not otherwise included in this set
+     * are cleared from the thread of execution for the duration of the
+     * action or task.</p>
+     *
+     * @param types types of thread context to capture and propagate.
+     * @return the same builder instance upon which this method is invoked.
+     */
+    ManagedExecutorBuilder propagated(String... types);
```
Emily-Jiang: This is type safe though. Is it possible to pass in ThreadContext type instead and then in ThreadContext to create instances for the contexts

njr-11: It should first be pointed out the builder pattern here exactly mirrors the CDI pattern for defining annotatively.  See the `propagated` method on https://github.com/eclipse/microprofile-concurrency/blob/master/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java, which is used as
`@Inject @ManagedExecutorConfig(propagated = { ThreadContext.SECURITY,  ThreadContext.APPLICATION)`
If this is going to be changed, it should be done on all of ManagedExecutorBuilder, ThreadContextBuilder, ManagedExecutorConfig, and ThreadContextConfig for consistency.
It would be wrong to supply a `ThreadContext` instance to the `ManagedExecutorBuilder.propagated` method because `ThreadContext` also allows for the specification of `unchanged` context which would be non-deterministic and behaviorally incorrect for CompletableFutures, undermining the purpose of this spec.
I'm open to some sort of enumerated type here rather than constants, but we would need to figure out how to do that in a way that is extensible (We want third party thread context providers to define their own context types that can be supplied here).
In any case, I would consider this design point a separate issue given that the changes under this pull are only copying what is already done by the existing `ManagedExecutorConfig`.

FroMage:
ok to test

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
njr-11: Issue is: https://github.com/eclipse/microprofile-concurrency/issues/19

FroMage:
Can we merge this please?

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+     * actions and async executor tasks that can be running at any given point
+     * in time. Async actions and tasks remain queued until capacity is available
+     * to execute them.</p>
+     *
+     * <p>The default value of <code>-1</code> indicates no upper bound,
+     * although practically, resource constraints of the system will apply.</p>
+     *
+     * @param max upper bound on async completion stage actions and executor tasks.
+     * @return the same builder instance upon which this method is invoked.
+     * @throws IllegalArgumentException if 0 or less than -1.
```
Emily-Jiang: it should be saying
if max is 0 or less than -1

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+     * <p>Establishes an upper bound on the number of async actions and async tasks
+     * that can be queued up for execution. Async actions and tasks are rejected
+     * if no space in the queue is available to accept them.</p>
+     *
+     * <p>The default value of <code>-1</code> indicates no upper bound,
+     * although practically, resource constraints of the system will apply.</p>
+     *
+     * @param max upper bound on async actions and tasks that can be queued.
+     * @return the same builder instance upon which this method is invoked.
+     * @throws IllegalArgumentException if 0 or less than -1.
```
Emily-Jiang: same as above comment

Issue #7, opened 2018-09-04
concurrency: support context propagator SPI


FroMage:
In order for mp-concurrency to replace [Reactive Contexts](https://reactiverse.io/reactive-contexts/) which is currently used by RESTEasy to provide context propagation, it would be very useful if there could be an SPI to declare custom context propagators.

For example, RxJava 1 and 2 both support a plugin system where it is relatively easy to implement automatic context propagation with a context API such as mp-concurrency and Reactive Contexts.

Supporting an SPI for context propagators is as simple as defining an interface with a single `setup` method, and an entry in `META-INF/services` to declare implementations.

With this in hand, mp-concurrency could provide two optional dependencies that provide implementations that hook into RxJava 1 and 2 in order to use mp-concurrency context propagation to propagate contexts automatically to all RxJava pipelines, rather than require users to implement it themselves or depend on a third-party library.

Copied from https://github.com/eclipse/microprofile-sandbox/issues/19

FroMage:
I'll try to re-explain this issue here in case my original explanation was not clear.

In RESTEasy and Reactive Contexts we take advantage of RxJava's plugin system to make context propagation automatic. For example the following code works out of the box:

```java
@GET
public Single<Hello> hello(){
 return Single.just("Hello")
  .delay(2, TimeUnit.SECONDS)
  .map(string -> string); // you can use your context here
}
```

This works because the `resteasy-rxjava2` module imports the `reactive-contexts-propagators-rxjava2` which implements the propagator SPI to register hooks into RxJava2 so that any callback created in an RxJava2 pipeline will capture the context and restore it in all subsequent listeners.

This is much better than requiring the user to use the context propagation API explicily, especially since context propagation is the better default.

For example, `CompletionStage`, which does not support plugins, requires users to do it explicitly by injecting a `ThreadContext` and calling `ctx.withContextCapture()` on _every_ `CompletionStage` they get.

Does this help?

FroMage:
This has been implemented by #48

PR #13, opened 2018-09-13
captureAllContext convenience method for usage of SPI by rxjava for Issue #7

njr-11:
Add a convenience method to the SPI that rxjava can use (see issue #7) to take advantage of the SPI without needing to write boilerplate code.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextProvider.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextProvider.java
@@ -18,7 +18,13 @@
  */
 package org.eclipse.microprofile.concurrent.spi;
 
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
+import java.util.ServiceLoader;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


njr-11:
Fully agree regarding not wanting to load the service providers every time.  I placed a TODO in the example code within this pull indicating that we could cache everything above that point to avoid that overhead, with the intention of adding it if we decide this pull otherwise solves the rxjava scenario.

Regarding your question about capture/clearing contexts, I assumed that you wanted all contexts captured for rxjava, which is what I currently have the captureAllContexts method doing, but if it requires the ability to specify exactly which to propagate vs leave unchanged, one idea would be to rename the method and update the signature to
captureContexts(props, propagatedContexts, unchangedContexts)

I'm not sure if you are looking for something like that or saying that we ought to go back to utilizing the ThreadContext API to solve the rxjava plugin scenario.

I suppose one question would be whether we are going to include the builders regardless of whether rxjava needs them.  If we will be including them for direct application usage without CDI regardless, and if they are also sufficient for the plugin, then we wouldn't need a ThreadContextProvider.capture[All]Context method.  There would certainly be advantages to that, in that the spec would consist entirely of interfaces rather than also including implementation for the static method.

njr-11:
This was determined not to be the best solution, so cancelling out.

PR #14, opened 2018-09-14
fix inconsistences between propagated and unchanged

### LINKED ISSUE #12: Questions about ThreadContextProvider and propagate/unchanged
- I think there's a discrepancy between the naming of `ThreadContextBuilder.propagate` and `ThreadContextBuilder.unchanged`. I think the first should either be `propagated` or the second be `ignore`.
- `ThreadContextBuilder` has `propagate`/`unchanged` but `ManagedExecutionBuilder` only has `context`, which I guess is the equivalent of `propagate`. The name should be coherent, and the functionality to set `unchanged` contexts too.
- The relation between `propagate` and `unchanged` sets, and the implied `cleared` set should be clarified:
-- For `propagate` we take all specified providers plus all the prerequisites, recursively. For `unchanged` the docs say to take all specified providers plus all those that have it as prerequisite, recursively. I wonder why the relation is reversed here? 
-- Once we have both sets, I assume the `cleared` set is the rest of the providers, but I suspect that we have to also make sure that none of the recursive prerequisites of the `cleared` set are within either of the `propagate` or `unchanged` set, right?

njr-11:
pull fixes #12

Renamed `context` on ManagedExecutorConfig to `propagated` in order to match the tense of `unchanged` and to be consistent with ThreadContextBuilder.  Updated docs accordingly.
Corrected explanation for implied prereqs for `unchanged` to match that of `propagated`.
Also removed a TODO from ThreadContext.ALL JavaDoc after placing the appropriate text in the `unchanged` JavaDoc which the TODO referred to. 

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextConfig.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextConfig.java
@@ -53,7 +53,7 @@ public @interface ThreadContextConfig {
      * inclusion of the prerequisites, even if not explicitly specified.</p>
      *
      * <p>Thread context types which are not otherwise included in this set or
-     * in the {@link #unchanged} set are removed from the thread of execution
+     * in the {@link #unchanged} set are cleared from the thread of execution
      * for the duration of the action or task.</p>
      *
      * <p>A <code>ThreadContext</code> must fail to inject if the same context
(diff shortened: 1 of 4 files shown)


FroMage:
LGTM. Please merge.

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextConfig.java
```diff
@@ -67,6 +67,14 @@
      * in that they are neither captured nor are they propagated or cleared
      * from thread(s) that execute the action or task.</p>
      *
+     * <p>Constants for specifying some of the core context types are provided
+     * on {@link ThreadContext}. Other thread context types must be defined
+     * by the specification that defines the context type or by a related
+     * MicroProfile specification. A <code>ThreadContext</code> must fail to
```
Emily-Jiang: fail with DefinitionException at application startup.

PR #15, opened 2018-09-17
ThreadContext method for obtaining a multiple-use context snapshot

### LINKED ISSUE #7: concurrency: support context propagator SPI

njr-11:
This pull adds the ThreadContext.withCurrentContext() method which returns a context snapshot executor, which can be used to contextualize essentially any operation.
contextShapshot = ThreadContext.withCurrentContext();
...
contextSnapshot.execute(() -> doSomethingThatRequiresContext());
contextSnapshot.execute(() -> doSomethingElseThatRequiresContext());

Pull helps with #7 

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
@@ -110,11 +110,11 @@ public interface ThreadContext {
     /**
      * <p>Creates an <code>Executor</code>that runs tasks on the same thread from which
      * <code>execute</code>is invoked but with context that is captured from the thread
-     * that invokes <code>withCurrentContext</code>.</p>
+     * that invokes <code>currentContextExecutor</code>.</p>
      *
      * <p>Example usage:</p>
      * <pre>
-     * <code>Executor contextSnapshot = threadContext.withCurrentContext();
+     * <code>Executor contextSnapshot = threadContext.currentContextExecutor();
      * ...
      * // from another thread, or after thread context has changed,
      * contextSnapshot.execute(() -> obj.doSomethingThatNeedsContext());
@@ -123,7 +123,7 @@ public interface ThreadContext {
      *
      * @return an executor that wraps the <code>execute</code> method with context.
      */
-    Executor withCurrentContext();
+    Executor currentContextExecutor();
 
     /**
      * <p>Wraps a <code>BiConsumer</code> with context that is captured from the thread that invokes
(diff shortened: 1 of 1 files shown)


njr-11:
@FroMage could you review this pull and add a comment if you believe it adequately addresses issue #7 which you opened.  Thanks!

FroMage:
Yes, that's useful, and can be used by plugins implementing #7 if we add it to the spec.

genie-microprofile:

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
```diff
@@ -106,6 +107,24 @@
...
+     * <code>Executor contextSnapshot = threadContext.withCurrentContext();
+     * ...
+     * // from another thread, or after thread context has changed,
+     * contextSnapshot.execute(() -> obj.doSomethingThatNeedsContext());
+     * contextSnapshot.execute(() -> doSomethingElseThatNeedsContext(x, y));
+     * </code></pre>
+     *
+     * @return an executor that wraps the <code>execute</code> method with context.
+     */
+    Executor withCurrentContext();
```
Emily-Jiang: ThreadContext.withCurrentContext() does not make much sense/

How about
ThreadContext.getExecutorWithCurrentContext()

FroMage: I disagree, I think that's too long a name and `withCurrentContext` is in line with the other wrapper functions already present here. They do exactly the same kind of thing too.

njr-11: I can see an object to `withCurrentContext` being that in other cases these methods accept a parameter that they wrap _with_ the current context, whereas in this case nothing is wrapped and instead a new executor instance is created.  Another idea for a name could be `currentContextExecutor` or just `contextualExecutor`.  Would either of these be acceptable to both of you?

FroMage: Yeah,that's fine.

FroMage:
Yes, LGTM.

PR #20, opened 2018-09-26
clarify error conditions for #18

### LINKED ISSUE #18: Clarify thread context dependency errors
We should clarify what happens in the following situations:

- more than one `ThreadContextProvider` has the same `threadContextType`
- a `ThreadContextProvider` lists a `prerequisite` which is missing (no `ThreadContextProvider` has that `threadContextType`)
- a `ThreadContextProvider` depends on itself via its `prerequisite` (directly or not)
- a user requests a thread context which is in both `propagated` and `unchanged` sets, directly or via `prerequisite`

The docs currently say "must fail to inject" in two of those cases, but that's too vague: 

- Should an error be generated?
- If yes: which kind?
- If yes: at what point?
- On `ThreadContext` build time, or at context _capture_ time or at context _set_ time?

I would suggest that we do generate an error (`IllegalStateException` or a custom one), as early as possible, such as on `ThreadContext` build time, so that no `ThreadContext` instance can be obtained that would have an erroneous set of contexts to propagate/clear/ignore.

njr-11:
pull fixes #18

Add clarification for error conditions, being consistent between CDI annotative config versus builder config.

For error conditions with thread context providers, this uses IllegalStateException for builders and DeploymentException for CDI.
For configuration errors made by the application, this uses IllegalArgumentException for builders and DefinitionException for CDI.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
@@ -70,8 +70,10 @@ public @interface ManagedExecutorConfig {
     /**
      * <p>Establishes an upper bound on the number of async completion stage
      * actions and async executor tasks that can be running at any given point
-     * in time. Async actions and tasks remain queued until capacity is available
-     * to execute them.</p>
+     * in time. There is no guarantee that async actions or tasks will start
+     * running immediately, even when the <code>maxAsync</code> constraint has
+     * not get been reached. Async actions and tasks remain queued until
+     * the <code>ManagedExecutor</code> starts executing them.</p>
      *
      * <p>The default value of <code>-1</code> indicates no upper bound,
      * although practically, resource constraints of the system will apply.</p>
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
@@ -32,6 +32,15 @@ import java.lang.annotation.Target;
  * ManagedExecutor executor;
  * ...
  * </code></pre>
+ *
+ * <p>A <code>ManagedExecutor</code> must fail to inject, raising
+ * <code>DeploymentException</code> on application startup,
+ * if the direct or indirect
+ * {@link org.eclipse.microprofile.concurrent.spi.ThreadContextProvider#getPrerequisites prerequisites}
+ * of a <code>ThreadContextProvider</code> are unsatisfied,
+ * or a provider has itself as a direct or indirect prerequisite,
+ * or if more than one provider provides the same thread context
+ * {@link org.eclipse.microprofile.concurrent.spi.ThreadContextProvider#getThreadContextType type}.
  */
 @Retention(RUNTIME)
 @Target(FIELD)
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
```diff
@@ -76,6 +89,10 @@
      *
      * <p>The default value of <code>-1</code> indicates no upper bound,
      * although practically, resource constraints of the system will apply.</p>
+     *
+     * <p>A <code>ManagedExecutor</code> must fail to inject, raising
+     * <code>DefinitionException</code> on application startup, if the
+     * <code>maxQueued</code> value is 0 or less than -1.
```
Azquelt: Why can't `maxQueued` be zero?

Intuitively, that would mean that if the maximum number of tasks are already executing, trying to submit a new one would result in an exception.

I note that this change is to match the doc on `ManagedExecutorBuilder`, so I can raise a new issue for this question instead if that's preferable.

njr-11: @Azquelt It would also mean that when you submit your very first task, if the system doesn't give you priority over everything thing else and start running that task immediately (if necessary even creating a new thread to do that), then your task or dependent stage fails because it cannot queue anything.  This is a bad practice that I don't want to use the spec to encourage it.  Inevitably, if 0 is an allowed value, users will try to configure it because at first glance it sounds nice (my work is really important - never queue it), and then end up with all sorts of unwanted failures in their application.  It would also disallow implementations from taking the approach of queuing every task and pulling from the queue as threads become available.  I cannot see adding a change that will lead to users following a bad practice and will be unnecessarily limiting upon implementations.

That said, I could see a case for instead allowing a 0 value for `maxAsync`.  A user might want to enforce that no one can use the *Async methods for dependent stages.  If this is something you were hoping for, I'm willing to allow the 0 value there instead.

FroMage:
Yes, LGTM.

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
njr-11: If we do this, let's use a separate issue, so that this pull can be merged.

Azquelt: I think so. When there's a spec you'll nail down what a running task (as it applies to `maxAsync`) is and what a queued task is.

I think my question comes out of not understanding your model of execution. As you described it, having `maxQueued` doesn't make sense.

However, it's important that these concepts are well defined and explained. Reading the description, I thought "if I have `maxAsync=10` and `maxQueued=10` then I can submit 20 tasks and none should be rejected. However, from your description that isn't true. Just because there's less than `maxAsync` currently running, doesn't mean a newly added task will start immediately.

njr-11: Thank you for clarifying further.  I think I understand now what the problem is.  You are reading the description of maxAsync as though it is a minimum constraint (tasks must start if less than this value) rather than as a maximum.  I have added a commit to update the maxAsync JavaDoc to explicitly state that it does not guarantee tasks will start immediately and to replace the language about running them subject to available capacity (which you probably interpreted as referring to the executor's maxAsync value rather than the system as a whole which was the intended meaning).

njr-11:
@FroMage would you mind adding a comment confirming that the additional commit that I added to clarify the JavaDoc to address the concern raised by @Azquelt is acceptable.  It looks like I now have sufficient access to merge the pull, but I want to be certain that the last commit was reviewed before doing so. (It was added after your prior review)

FroMage:
Yes, LGTM too.

PR #22, opened 2018-09-28
newIncompleteFuture method for ManagedExecutor interface for #21

### LINKED ISSUE #21: ManagedExecutor.newIncompleteFuture
One thing that we are missing on ManagedExecutor is an equivalent for the ability to construct a new instance of CompletableFuture which is not completed and chain dependent stages, later completing the original CompletableFuture.  A workaround for this is requesting an already-completed completable future, 
`managedExecutor.completedFuture(unusedResultOfCorrectType).newIncompleteFuture();`
However, this pattern seems like it could be common enough that a convenience method would be valuable,
`CompletableFuture<Integer> cf = managedExecutor.newIncompleteFuture(Integer.class)`

njr-11:
Add newIncompleteFuture method to ManagedExecutor so that users have a way of constructing a CompletableFuture that they can complete at a later time, the same as they can do with the Java SE CompletableFuture implementation.

pull fixes #21

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
@@ -122,6 +122,18 @@ public interface ManagedExecutor extends ExecutorService {
     <U> CompletionStage<U> failedStage(Throwable ex);
 
     /**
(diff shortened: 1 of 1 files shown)


genie-microprofile:

FroMage:
LGTM

PR #23, opened 2018-10-01
SPI to automatically plug in context propagators, for issue  #7

### LINKED ISSUE #7: concurrency: support context propagator SPI

njr-11:
pull fixes #7

Use notification of successful ConcurrencyProvider registration as the invocation point for context propagator hooks.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyProvider.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyProvider.java
@@ -63,10 +63,21 @@ public interface ConcurrencyProvider {
     ThreadContextBuilder newThreadContextBuilder(); 
 
     /**
-     * Allows the container to register the <code>ConcurrencyProvider</code>
+     * <p>Allows the container to register the <code>ConcurrencyProvider</code>
      * implementation. At most one implementation can be registered at any
      * given point in time. In order to register a different implementation,
-     * the container must first unregister its previous implementation.
+     * the container must first unregister its previous implementation.</p>
+     *
+     * <p>Upon successful registration of the <code>ConcurrencyProvider</code>,
+     * the MicroProfile Concurrency implementation must subsequently query the
+     * container's <code>ServiceLoader</code> for all
+     * <code>ConcurrencyProviderRegistrationListener</code>s and send the
+     * <code>providerRegistered</code> notification to each.
+     * <code>ConcurrencyProviderRegistrationListener</code> is an SPI level
+     * interface. It must not be supplied directly within applications because
+     * application classes might not be available on the
+     * <code>ServiceLoader</code> within the scope of the provider
+     * registration.</p>
      *
      * @param provider the provider implementation to register.
      * @throws IllegalStateException if an implementation is already registered.
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyProviderRegistrationListener.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyProviderRegistrationListener.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


genie-microprofile:

njr-11:
@FroMage please review the changes under this pull as a solution to issue #7 which you opened to have a convenient way in the spec for third parties to register hooks for context propagators.  When I looked into where in the life cycle this would fit in, it made sense that it should happen as soon as possible after the ConcurrencyProvider is registered, so I named the interfaces more generically to represent their alignment with that pattern.  Let me know if you have better names or if you would prefer that we use the exact names/signatures from your examples.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyProvider.java
```diff
@@ -63,10 +63,21 @@ public static ConcurrencyProvider instance() {
...
-     * Allows the container to register the <code>ConcurrencyProvider</code>
+     * <p>Allows the container to register the <code>ConcurrencyProvider</code>
      * implementation. At most one implementation can be registered at any
      * given point in time. In order to register a different implementation,
-     * the container must first unregister its previous implementation.
+     * the container must first unregister its previous implementation.</p>
+     *
+     * <p>Upon successful registration of the <code>ConcurrencyProvider</code>,
+     * the MicroProfile Concurrency implementation must subsequently query the
+     * container's <code>ServiceLoader</code> for all
```
manovotn: Is it then legit to have more than one `ConcurrencyProvider` in a running application?

FroMage: No it's explicitly forbidden.

FroMage:
In smallrye-concurrency I have made it take a `ThreadContext` parameter, because propagators are per-module, like context providers. If we load the set of propagators in the `ContextProvider` we can't have that set be per-module.

The problem is that I can't reproduce the per-module issue in RESTEasy's CI ATM, for some reason I can't explain, so I probably will have to manually write a test to reproduce the problematic situation before I can actually validate these sorts of issues.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyProvider.java
manovotn: Depends, reading on a `register()` method, it is forbidden there.
See [here](https://github.com/eclipse/microprofile-concurrency/blob/master/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyProvider.java#L100).

And I think it doesn't make much sense to have more than one there.

njr-11: Fully agree it is explicitly forbidden to have more than one ConcurrencyProvider.  Just let me know if you see anywhere that suggests otherwise, and I'll be happy to get it updated.

njr-11:
> In smallrye-concurrency I have made it take a `ThreadContext` parameter, because propagators are per-module, like context providers. If we load the set of propagators in the `ContextProvider` we can't have that set be per-module.
> 
> The problem is that I can't reproduce the per-module issue in RESTEasy's CI ATM, for some reason I can't explain, so I probably will have to manually write a test to reproduce the problematic situation before I can actually validate these sorts of issues.

The reason that `providerRegistered(ConcurrencyProvider provider)` has a ConcurrencyProvider parameter rather than ThreadContext is to ensure the propagator has the opportunity to determine which thread context types are propagated rather than being stuck with a set of defaults.

FroMage:
> The reason that providerRegistered(ConcurrencyProvider provider) has a ConcurrencyProvider parameter rather than ThreadContext is to ensure the propagator has the opportunity to determine which thread context types are propagated rather than being stuck with a set of defaults.

Good point.

FroMage:
I believe that with #30 we can turn that method into `providerRegistered(ConcurrencyManager)` and then we're good because the set of propagators will be tied with the set of providers, with a manager per classloader.

FroMage:
I've done the modification I proposed in #48.

PR #25, opened 2018-10-09
Start of a TCK

FroMage:
This is a setup copied heavily from MP-Config, and could work as the basis of our TCK. It's using JUnit and Arquillian.

I had to change the way ConcurrencyProvider was registered, though, by using an SPI, again taking inspiration from MP-Config. I'll open an issue to discuss that.

### DIFF
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -56,7 +56,7 @@
         <dependency>
             <groupId>javax.enterprise</groupId>
             <artifactId>cdi-api</artifactId>
-            <version>1.2</version>
+            <version>2.0.SP1</version>
             <scope>provided</scope>
         </dependency>
 
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -1,0 +1,68 @@
+<?xml version="1.0"?>
+<!--
+    Licensed under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+     http://www.apache.org/licenses/LICENSE-2.0
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+-->
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+    xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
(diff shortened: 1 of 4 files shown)


njr-11:
This pull also fixes #26 which is an issue to add the ability to register the ConcurrencyProvider via the ServiceLoader

FroMage:
Note that https://github.com/smallrye/smallrye-concurrency/tree/tck/tck has an example of how to invoke the TCK (from the IDE via the `SmallRyeTckSuite` class or the CLI with `mvn test`)

tck/src/main/java/org/microprofile/concurrency/tck/TckTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(Arquillian.class)
+public class TckTest {
+
+    @Deployment
+    public static WebArchive createDeployment() {
```
manovotn: Any specific reason why the deployment is `WebArchive`?
Because as it is written now, this will result in duplicate `TckTest` class in the archive. E.g. the structure is as follows:

someTestArchive.war
+-- _WEB-INF
|   +-- classes/org.micro..../TckTest.class
|   +-- lib/createdJarArchive.jar
|         +-- META-INF/beans.xml
|         +-- org.micro..../TckTest.class

I am not an Arq. expert, but I think the reason is that Arquillian automatically adds the test class into the root archive.
The only exception is when it is run in a [client mode](https://docs.jboss.org/author/display/ARQ/Test+run+modes).

FroMage: Frankly I don't know, I can try to change it and see if it still works. I copied this set up from mp-config.

manovotn: Well, you can either go with just jar, where the test class will also be automatically added or make it run in client mode.
From the TCK runner in smallrye I saw that you are using weld-embedded anyway.
That's basically starting Weld SE container and assumes "se-like structure" which means flat one (at least I think it does :)). If that's the case, having WAR has no effect versus JAR.

manovotn: On second thought, WAR is probably what users use more often.
So we might just fix the duplicate class issue and leave it at that.

tck/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+            <groupId>org.eclipse.microprofile.concurrency</groupId>
+            <artifactId>microprofile-concurrency-api</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>javax.enterprise</groupId>
+            <artifactId>cdi-api</artifactId>
+            <version>1.2</version>
```
manovotn: We should be going for 2.0 version (2.0.SP1 to be precise).

FroMage:
I'll merge this since it was approuved. I just separated the SPI part away and bumped CDI.

Issue #24, opened 2018-10-05
What to do when user specifies a type of thread context that isn't available


njr-11:
It is possible for a user to either directly or indirectly request a type of thread context to be propagated that isn't actually available.  The spec should clarify what happens in this case, so that implementations do not differ.

For example, with the following, assume that a CDI implementation is either not provided by the product or the part of the product that provides CDI is disabled,

```
@Inject @ManagedExecutorConfig(propagated = { ThreadContext.APPLICATION, ThreadContext.SECURITY, ThreadContext.CDI })
ManagedExecutor executor1;

@Inject @ManagedExecutorConfig // the default is the same as above
ManagedExecutor executor2;
```

One option is to simply ignore the types that aren't provided.  The idea here being that if a feature such as CDI isn't provided/usable in the product, that it doesn't matter that we cannot propagate context for it, and we shouldn't trouble the user with an error.  However, this overlooks the possibility that a feature like CDI is provided/usable but the particular CDI implementation is one that didn't contribute a ThreadContextProvider implementation.  Adding a method like managedExecutor/threadContext.propagates(type) to indicate what will actually propagate could help here, but that is also more clutter to the programming model.  Probably the biggest advantage of ignoring types that aren't provided is that the default value of `propagated` will always be usable without causing errors.  Could also include a one-time warning for ignored context.

The other option is to reject the inject/build operation when a thread context type is not provided.  With this, the user is guaranteed that any type they specific will be propagated, and they will be forced to confront errors for types the specify which are not present/enabled.  This will be an advantage when it means detecting a context type for which the user didn't realize a ThreadContextProvider was missing.  If we do this, the defaults for `propagated` will need to be switched, probably to ALL, to avoid causing errors on the basic path where the user has not bothered to configure which specific types.

Please feel free to add further discussion.  On the MP Concurrency call where this was brought up, it sounded like more people were in favor of the second (reject) option.

njr-11:
Per the switch to make ALL_REMAINING be the default for propagated types under #28, we no longer have a situation where default propagated context could be unavailable.  This means we can now choose the option where we reject the inject/build operation when specified context to propagate is not available.

PR #28, opened 2018-10-16
define defaults for context propagation

### LINKED ISSUE #27: Determine which contexts should propagate by default
At first glance, propagating all available context seems like a nice, consistent default to have that will be intuitive for users.  However, certain context types, namely transaction context, can have some very undesirable consequences that will be inconvenient for users to code around, and, if unaware of it, could lead to data corruption (due to operating under an unexpected transaction) and deadlocks.

It is my own opinion that we should continue the default behavior established by the EE Concurrency spec in by default, establishing a cleared/empty transaction context for dependent stages, within which the user is free to begin and commit/roll back their own transactions if they wish to.  This seems to me by far the most likely and desirable pattern in which users will wish to use transactions in combination with CompletableFuture and/or other async tasks.  Propagating a transaction across threads seems to me like an advanced pattern that, when the user so desires, it will be natural to explicitly request this behavior (propagate=ThreadContext.TRANSACTION).

The flip side of the argument, if I am representing it correctly, is that because this capability isn't currently possible, we don't know how popular it will become.  Users might come to expect the consistence of having all thread context types propagated, including transaction context.  And of course, if the user expected it to be propagated and it wasn't, the very same arguments around preventing data corruption apply equally for this case as well.

njr-11:
fixes #27

Add a constant for context propagation defaults per discussion under #27

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
@@ -46,15 +46,41 @@ import java.lang.annotation.Target;
 @Target(FIELD)
 public @interface ManagedExecutorConfig {
     /**
+     * <p>Defines the set of thread context types to clear from the thread
+     * where the action or task executes. The previous context is resumed
+     * on the thread after the action or task ends.</p>
+     *
+     * <p>By default, the transaction context is cleared/suspended from
+     * the execution thread so that actions and tasks can start and
+     * end transactions of their choosing, to independently perform their
+     * own transactional work, as needed.</p>
+     *
+     * <p>Constants for specifying some of the core context types are provided
+     * on {@link ThreadContext}. Other thread context types must be defined
+     * by the specification that defines the context type or by a related
+     * MicroProfile specification.</p>
+     *
+     * <p>Inclusion of a thread context type with prerequisites implies
+     * inclusion of the prerequisites, even if not explicitly specified.</p>
+     *
+     * <p>A <code>ManagedExecutor</code> must fail to inject, raising
+     * <code>DefinitionException</code> on application startup, if the same
+     * context type is implicitly or explicitly included in this set
+     * as well as in the set specified by {@link #propagated}.</p>
+     */
+    String[] cleared() default { ThreadContext.TRANSACTION };
+
+    /**
      * <p>Defines the set of thread context types to capture from the thread
      * that creates a dependent stage (or that submits a task) and which to
      * propagate to the thread where the action or task executes.</p>
      *
      * <p>The default set of thread context types is
-     * {@link ThreadContext#DEFAULTS}, which includes all available
+     * {@link ThreadContext#ALL_OTHER}, which includes all available
      * thread context types that support capture and propagation to other
-     * threads, except for {@link ThreadContext#TRANSACTION} context, which
-     * is instead cleared (suspended) from the thread that runs the action or
+     * threads, except for those that are explicitly {@link cleared},
+     * which, by default is {@link ThreadContext#TRANSACTION} context,
+     * in which case is suspended from the thread that runs the action or
      * task.</p>
      *
      * <p>Constants for specifying some of the core context types are provided
@@ -68,8 +94,13 @@ public @interface ManagedExecutorConfig {
      * <p>Thread context types which are not otherwise included in this set
      * are cleared from the thread of execution for the duration of the
      * action or task.</p>
+     *
+     * <p>A <code>ManagedExecutor</code> must fail to inject, raising
+     * <code>DefinitionException</code> on application startup, if the same
+     * context type is implicitly or explicitly included in this set
+     * as well as in the set specified by {@link #cleared}.</p>
      */
-    String[] propagated() default { ThreadContext.DEFAULTS };
+    String[] propagated() default { ThreadContext.ALL_OTHER };
 
     /**
      * <p>Establishes an upper bound on the number of async completion stage
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
@@ -50,8 +50,12 @@ public @interface ManagedExecutorConfig {
      * that creates a dependent stage (or that submits a task) and which to
      * propagate to the thread where the action or task executes.</p>
      *
-     * <p>The default set of thread context types is those required by the
-     * EE Concurrency spec, plus CDI.</p>
+     * <p>The default set of thread context types is
+     * {@link ThreadContext#DEFAULTS}, which includes all available
+     * thread context types that support capture and propagation to other
+     * threads, except for {@link ThreadContext#TRANSACTION} context, which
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
```diff
@@ -65,7 +69,7 @@
      * are cleared from the thread of execution for the duration of the
      * action or task.</p>
      */
-    String[] propagated() default { ThreadContext.APPLICATION, ThreadContext.CDI, ThreadContext.SECURITY };
+    String[] propagated() default { ThreadContext.DEFAULTS };
```
tomjenkinson: In my opinion this is a bit confusing (it to remove the explicit set of defaults in the Java code and rely on readers to the Javadoc). It seems to be similar opinion of author of comment https://github.com/eclipse/microprofile-concurrency/issues/27#issuecomment-430120816 @manovotn

manovotn: Yes, but the PR now basically states that all of the context excepting transaction are propagated.
That's IMO good - what I didn't like was the previous notion of leaving default undefined and based on implementation.

njr-11: @tomjenkinson there are a couple of problems that are causing us to move away from the approach of enumerating an exact list of individual contexts as the default.

1. it makes the default ManagedExecutor/ThreadContext unusable if any one of the types listed isn't supported by the product or has been intentionally disabled by the user.  For example, if a product doesn't include CDI or the user turns off CDI, then the basic @Inject ManagedExecutor will be unusable per the direction we are headed under #24.
2. It makes for a suboptimal user experience when a container or third party provides any additional context types.  The user would miss out on propagation of these types by default.

Possibly, DEFAULTS isn't the best name for this, and I'm open to improvements that anyone has to offer.  What it really means to say is either
"all available contexts except for transaction context, which is instead suspended/cleared"
or
"all available contexts which say they should be default where transaction context must claim it is not default and is instead suspended/cleared".
There is an ongoing discussion between the above options on the meaning for DEFAULTS.
We also have the ALL constant for all available contexts _including_ transactions, which could be trivially replaced by (DEFAULTS, TRANSACTION) if the former meaning is chosen.

Based on your comment, I can guess that you would favor the former meaning here.  In that case, would it be helpful to rename DEFAULTS to ALL_EXCEPT_TRANSACTION ?

tomjenkinson: Given what you said it sounds a bit more like a blacklist rather than a whitelist (as those in the whitelist my not be available for the reasons you mentioned).
Maybe:
`String[] defaultNotPropagated() { ThreadContext.TRANSACTION }`

njr-11: @tomjenkinson The name `defaultNotPropagated` doesn't make sense to me (a default isn't a configurable attribute that a user sets to various values), but I understand the general idea here, and it is a good idea.

There are three categories of context propagation (propagated, unchanged, cleared), but only the first two of these are currently configurable, with the third being implied by absence of a context type in the other sets.  However, we could do this differently and allow the explicit configuration of all three sets.  It would be redundant, but more straightforward, and I think probably worth it.

In this case, we would define the ALL constant to mean all available context types that aren't otherwise individually listed elsewhere.  Would ALL_OTHER be a more accurate name? Does anyone have preferences on that?

So, for ThreadContext, we would have:

```
String[] propagated() default { ThreadContext.ALL };
String[] cleared() default { ThreadContext.TRANSACTION };
```

And for ManagedExecutor, we would have:

```
String[] propagated() default { ThreadContext.ALL };
String[] cleared() default { ThreadContext.TRANSACTION };
```
* Note that `unchanged` is intentionally forbidden on ManagedExecutor because it would introduce the same non-deterministic, unsafe, unpredictable behavior that we wrote the spec to solve.

There would no longer be a need for a DEFAULTS constant because it would already be the default.

tomjenkinson: It would need to be clear in the text of the specification that cleared overrides propagated but yes. It some ways would seem to be a bit like how the excludes annotation is used in the Ant fileset example: https://ant.apache.org/manual/Types/fileset.html

`<fileset dir="${server.src}" casesensitive="yes">
  <include name="**/*.java"/>
  <exclude name="**/*Test*"/>
</fileset>`

njr-11: It isn't that cleared overrides propagated.  Consider the opposite example.  I think this just made the case for not naming it ALL and calling it ALL_OTHER instead,

```
@ThreadContextConfig(propagated = ThreadContext.APPLICATION,
                     cleared = ThreadContext.ALL_OTHER)

@ThreadContextConfig(cleared = TRANSACTION,
                     propagated = ThreadContext.ALL_OTHER)
```

tomjenkinson: I think it may be use the way I expressed override. If the first example means propagate everything but ThreadContext.APPLICATION and the second one is to propagate everything but TRANSACTION it looks good to me.

tomjenkinson: I edited my previous comment above^

njr-11:
@tomjenkinson Thanks for your comments. I think they have resulted in improving the design and making the configuration more intuitive.  I have added a commit that covers we discussed in your review comment thread.  Please review again. Thanks!
@manovotn and @FroMage please review as well.

njr-11:
On the MP Concurrency call we thought of ALL_REMAINING as a better name than ALL_OTHER.  We also decided that we will leave this pull open for several days in case anyone thinks of better names to consider.  After that, I'll add another commit that updates to the chosen name.

To recap the discussion, there are 3 ways of dealing with thread context (all configurable).
Thread context can be **propagated** from the requesting thread to the thread that runs the operation.
Thread context can be **cleared** from thread that executes the action.
Thread context can remain **unchanged** on the thread that executes the action. (This one is only applicable when using ThreadContext to run actions inline, not when using a ManagedExecutor)

Let's say that someone wants a ThreadContext instance that clears all context when running actions, except for Transaction context, which it wants unchanged on the execution thread, and Security context, which it wants propagated from the requesting thread.
Then they would configure,
```
@Inject @ThreadContextConfig(propagated=ThreadContext.SECURITY,
                             unchanged=ThreadContext.TRANSACTION,
                             cleared=ThreadContext.ALL_REMAINING)
ThreadContext threadCtx;
```

So this last question is just over whether name ALL_REMAINING is good and sufficiently intuitive for describing this, or whether someone thinks of a better name for it.

njr-11:
I haven't seen any suggestions come through for better names, so I have added a commit updating the constant name from ALL_OTHER to ALL_REMAINING as discussed on the MP Concurrency call last Friday.   I have also rebased the commits onto the latest from master, and included a couple of other minor JavaDoc fixes that I missed in the original pull.

@manovotn or @tomjenkinson could you approve this latest commit?

Issue #5, opened 2018-09-04
concurrency: support per-application SPI providers


FroMage:
In order for mp-concurrency to replace [Reactive Contexts](https://reactiverse.io/reactive-contexts/) which is currently used by RESTEasy to provide context propagation, further to #16 we need to be able to declare providers per application, and not just globally.

When integrating Reactive Contexts in RESTEasy we noticed that in WildFly we had issues because it uses a modular class loader (JBoss Modules) and supports multiple independent deployments of applications.

In this particular case the WildFly distrib was loaded with RESTEasy and Reactive Contexts, and then we had two independent user applications (two `.war` files) that each had their own sets of context providers.

Having a global list of context providers held as a static variable did not work: we needed to have a per-deployment list of context providers.

On deployment, RESTEasy instantiates a new Reactive Contexts subsystem, and stores it in a thread-local (the assumption being that every RESTEasy deployment is tied initially to a separate thread). From there, every thread which serves RESTEasy request has that thread-local propagated and it is used to look up the set of context providers that participate in context propagation.

Copied from https://github.com/eclipse/microprofile-sandbox/issues/17

njr-11:
Could this be addressed by documenting a requirement that ContextProviders are loaded using the classloader that is present on the thread at the point of injection (if CDI is used) or the point of construction (if builder approach is used) of the ThreadContext/ManagedExecutor instance ?

FroMage:
Well, it also depends how the ThreadContext is initialised. For ReactiveContexts in RESTEasy it's initialised at startup, and kept in RESTEasy's context (the irony). For each request it will then be set to the current thread's ReactiveContext, so that context capturing works.

The big problem was that we need context propagation for RxJava, which only supports a single hook, defined in a static. That hook (context propagator) needs to obtain the current context to capture/restore it. I don't think it can use injection, and it can't build a new context, so it has to look it up for the current request, which is why a thread-local works. After that it will let the context handle capturing/propagation so we don't need the thread-local anymore. But it's the initial lookup that has to be per-request. I suppose it could be per-ClassLoader, but still that association has to be kept somewhere so RxJava's context propagator can find it.

FroMage:
I believe #30 solves this issue.

njr-11:
Can this be closed out with delivery of pull #30 ?

FroMage:
Yes

PR #30, opened 2018-10-19
Support per-classloader configuration and specify when to load the ThreadContextProviders

FroMage:
I've finally been able to reproduce the issues related to having multiple sets of `ThreadContextProvider` and `ThreadContextPropagator` in separate applications (different wars with different context classloaders), in a container which provides MP-Concurrency (in a parent classloader).

The proposed fix is to copy the MP-Config configuration by introducing the concept of `ConcurrencyManager` which is the object we get/create out of the `ConcurrencyProvider`, whose role is to load all the sets of `ThreadContextProvider` and `ThreadContextPropagator` (not in this PR) for a given `ClassLoader`. A `ConcurrencyManager` will therefore have its own set of providers and be able to create new `ThreadContextBuilder` and `ManagedExecutorBuilder` instances for those providers.

Using `ConcurrencyProvider.getConcurrencyManager()` will lazily create (and memoize) a `ConcurrencyManager` for the current thread-context `ClassLoader`, which should work for most application types. Containers which can unload applications can call `ConcurrencyProvider.unregisterConcurrencyManager` to clear the cache. 

Additionally we now have a `ConcurrencyManagerBuilder` that lets us manually specify if the `ThreadContextProvider` should be service-loaded or manually loaded, and which `ClassLoader` to use.

Again, all this is copied from MP-Config, and there's a good reason it all went in there. I have an implementation of this all in SmallRye Concurrency which validates that these additions solve all the concerns I had regarding multiple applications, and will allow me to close several related issues such as #5 and #7.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyManagerBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyManagerBuilder.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.concurrent.spi;
+
+/**
+ * Use this class to configure instances of {@link ConcurrencyManager}.
+ */
+public interface ConcurrencyManagerBuilder {
+    
+    /**
+     * Use the specified {@link ThreadContextProvider} instances.
+     * @param providers the {@link ThreadContextProvider} instances to use.
+     * @return this builder
+     */
+    public ConcurrencyManagerBuilder withThreadContextProviders(ThreadContextProvider... providers);
+
+    /**
+     * Load all discoverable {@link ThreadContextProvider} instances via the {@link ServiceLoader}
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyManagerBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public ConcurrencyManagerBuilder withThreadContextProviders(ThreadContextProvider... providers);
+
+    /**
+     * Load all discoverable {@link ThreadContextProvider} instances via the {@link ServiceLoader}
+     * mechanism on the current thread-context {@link ClassLoader} (unless overridden by {@link #forClassLoader(ClassLoader)}).
+     * 
+     * @return this builder
+     */
+    public ConcurrencyManagerBuilder addDiscoveredThreadContextProviders();
```
njr-11: A minor comment on the builder - the presence of addDiscoveredThreadContextProviders without any parameters on a builder interface is a bit awkward.  Given this and the interaction between the two, we should consider if addDiscoveredThreadContextProviders and forClassLoader ought to be collapsed into a single builder method, such as,
```
withThreadContextProvidersFor(ClassLoader classLoader);
```

njr-11:
Overall, the concepts in this pull all look good.  This is exactly how I would expect it to work with loading providers based on the thread context class loader.  Although I'm wondering if these interfaces, which seem entirely something that the concurrency provider itself will both provide as well as use, actually need to be defined as spec classes rather than just being part of the concurrency provider's own implementation.  There is probably some subtle detail here that I'm not thinking about, and we can talk about it on the upcoming MP Concurrency call.  Overall I don't see anything objectionable about putting these SPI interfaces in the spec, just wanting to be sure that it is truly necessary.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ConcurrencyManagerBuilder.java
FroMage: I think it's meant to not repeat passing the `classLoader` for the rest of the providers, like when we add `withThreadContextPropagators()` and possibly more in the future.

FroMage:
What convinced me was that with this you can make containers implementation-agnostic because they can get/create and clear `ConcurrencyManager` instances for each application. Otherwise they would have to use the implementation-specific API and be tied to a specific implementation.

I also suspec the builder will make writing the TCK easier.

PR #35, opened 2018-10-25
fix javadoc to remove prerequisite context

### LINKED ISSUE #34: JavaDoc for builders/config has language around prereqs that prevents granular specification of thread context types
JavaDoc for the ThreadContextBuilder/Config and ManagedExecutorBuilder/Config methods currently include language around prerequisite context types which assumes that prerequisite types must always follow the same category of thread context propagation (cleared/propagated/unchanged),

```
     * <p>Inclusion of a thread context type with prerequisites implies
     * inclusion of the prerequisites, in that the prequisites are
     * considered *** as well, even if not explicitly specified.</p>
```

At first glance, this might seem nice, but it is actually overly limiting and creates a lot of unnecessary complexity for implementations  as well as for users trying to interpret the behavior they will end up with.

As an example, let's say that a CDI thread  context provider implements its getPrerequisites method to return ThreadContext.APPLICATION.  The intent here (and consistent with the documentation of ThreadContextProvider.getPrerequistes) would be for application context to be established on the thread before the container requests that CDI context be established on the thread.  The idea of prerequisite context was intended as a way to establish ordering.

However, with the additional language that is currently in the Builders/Config, it gets another meaning as well, by also implying propagation, clearing, or unchangedness.  Consider what this means if a user configures,
```@ThreadContextConfig(cleared=ThreadContext.CDI, propagated=ThreadContext.APPLICATION)```
This seems like a perfectly reasonable combination to configure, but the current language in the Builders/Config JavaDoc forces it to be an error due to the fact that APPLICATION ends up implicitly in one category and explicitly in another.

One solution to this would be to remove the language in the JavaDoc around establishing implicit prerequisites.  This will allow the user to specify exactly what they want without interference from the container/thread context provider implying additional types of context that the user didn't want or expect.  This would be the simplest and cleanest solution, although it would permit users to configure erroneous combinations that aren't detected by the builder implementation, forcing the ThreadContextProvider implementations themselves to detect and raise the errors.

If we wanted to attempt to address this at the level of the builder, then we need to first come up with some rules on how to interpret the many possible combinations.  Here is one possibility:
- It is valid to clear a context type that has prerequisites even if its prerequisites are not cleared.
- It is valid for a type and its prereqs to be applied in same way (for example, all propagated).
- It is not valid for a type and its prereqs to be applied in different ways.
- cleared does not imply prereqs are cleared.
- propagated implies prereqs are propagated.
- unchanged implies prereqs are unchanged.
- Prereqs of explicitly specified types are inferred before applying ALL_REMAINING.

Given the above, we could create the following truth table:
```
+------+---------------------------+---------------------------+---------------------------+
|valid?| cleared                   | propagated                | unchanged                 |
+------+---------------------------+---------------------------+---------------------------+
|   Y  | CDI,APPLICATION           |                           |                           |
|   Y  | CDI,ALL_REMAINING         |                           |                           |
|   Y  | CDI                       | APPLICATION               |                           |
|   Y  | CDI                       | ALL_REMAINING             |                           |
|   Y  | CDI                       |                           | APPLICATION               |
|   Y  | CDI                       |                           | ALL_REMAINING             |
|   N  | APPLICATION               | CDI                       |                           |
|   N  | APPLICATION               | ALL_REMAINING             |                           |
|   N  | APPLICATION               |                           | CDI                       |
|   N  | APPLICATION               |                           | ALL_REMAINING             |
|   Y  | APPLICATION,ALL_REMAINING |                           |                           |
|   y  | ALL_REMAINING             | CDI                       |                           |
|   Y  | ALL_REMAINING             | APPLICATION               |                           |
|   y  | ALL_REMAINING             |                           | CDI                       |
|   Y  | ALL_REMAINING             |                           | APPLICATION               |
|   Y  |                           | CDI,APPLICATION           |                           |
|   Y  |                           | CDI,ALL_REMAINING         |                           |
|   N  |                           | CDI                       | APPLICATION               |
|   y  |                           | CDI                       | ALL_REMAINING             |
|   Y  |                           | APPLICATION,ALL_REMAINING |                           |
|   N  |                           | APPLICATION               | CDI                       |
|   N  |                           | APPLICATION               | ALL_REMAINING             |
|   y  |                           | ALL_REMAINING             | CDI                       |
|   N  |                           | ALL_REMAINING             | APPLICATION               |
|   Y  |                           |                           | CDI,APPLICATION           |
|   Y  |                           |                           | CDI,ALL_REMAINING         |
|   Y  |                           |                           | APPLICATION,ALL_REMAINING |
+------+---------------------------+---------------------------+---------------------------+
```
But the above introduces a lot of complexity, both in the user and implementation needing to understand all of the rules.  Is it worth it?

njr-11:
pull fixes #34

Update JavaDoc for prerequisites to clarify that when providers declare context types as prerequisite, this helps determine the order in which context types are applied to the execution thread, but it does not create implicit propagation/clearing/unchanging of the prereq types when the thread context type that declares them prerequisite is configured to propagate/clear/remain unchanged.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -176,6 +176,5 @@ With this model, the provider of thread context implements the
 org.eclipse.microprofile.concurrent.spi.ThreadContextProvider
 interface and packages it in a way that makes it available to the
 ServiceLoader. ThreadContextProvider identifies the thread context
-type and any prerequisite thread context types that it has,
-and it allows for capturing snapshots of thread context
-as well as for applying empty/default context to threads.
+type and provides a way to capture snapshots of thread context
+as well as for applying empty/cleared context to threads.
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
@@ -50,11 +50,6 @@ public interface ManagedExecutorBuilder {
      *         {@link #cleared} set and the {@link #propagated} set</li>
      *         <li>if a thread context type that is configured to be
      *         {@link #cleared} or {@link #propagated} is unavailable</li>
-     *         <li>if the direct or indirect
-     *         {@link org.eclipse.microprofile.concurrent.spi.ThreadContextProvider#getPrerequisites prerequisites}
-     *         of a <code>ThreadContextProvider</code> are unsatisfied</li>
-     *         <li>if a <code>ThreadContextProvider</code> has a direct or
-     *         indirect prerequisite on itself</li>
      *         <li>if more than one provider provides the same thread context
      *         {@link org.eclipse.microprofile.concurrent.spi.ThreadContextProvider#getThreadContextType type}
      *         </li>
@@ -80,9 +75,6 @@ public interface ManagedExecutorBuilder {
      * on {@link ThreadContext}. Other thread context types must be defined
      * by the specification that defines the context type or by a related
      * MicroProfile specification.</p>
-     *
-     * <p>Inclusion of a thread context type with prerequisites implies
-     * inclusion of the prerequisites, even if not explicitly specified.</p>
      *
      * @param types types of thread context to clear from threads that run
      *        actions and tasks.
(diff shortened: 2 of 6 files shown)


njr-11:
@FroMage could you review these changes to remove the prerequisite context?  I first confirmed with Matej that CDI doesn't need it, and I agree it's best to remove it entirely and only bring it back if/when the need arises.

FroMage:
Yes, agreed.

PR #36, opened 2018-10-26
require build/inject to fail when user configures an unavailable context type

### LINKED ISSUE #24: What to do when user specifies a type of thread context that isn't available

njr-11:
As discussed in the issue that this fixes #24 , this pull updates JavaDoc to require that an error be raised on ManagedExecutorBuilder.build and ThreadContextBuilder.build when a specified type of thread context is not available.  And it similarly updates JavaDoc on ManagedExecutorConfig/ThreadContextConfig to similarly require that injection fail when a specified type of thread context is not available.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
@@ -44,9 +44,11 @@ public interface ManagedExecutorBuilder {
      * instances.</p>
      *
      * @return new instance of <code>ManagedExecutor</code>.
-     * @throws IllegalArgumentException if the same thread context type is
-     *         present in, or inferred by, both the {@link #cleared} set
-     *         and the {@link #propagated} set.
+     * @throws IllegalArgumentException if a thread context type that is
+     *         configured to be {@link #cleared} or {@link #propagated} is
+     *         unavailable. Also raises IllegalArgumentException if one or
+     *         more of the same context types appear in both the
+     *         {@link #cleared} set and the {@link #propagated} set.
      * @throws IllegalStateException if the direct or indirect
      *         {@link org.eclipse.microprofile.concurrent.spi.ThreadContextProvider#getPrerequisites prerequisites}
      *         of a <code>ThreadContextProvider</code> are unsatisfied,
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
@@ -44,17 +44,21 @@ public interface ManagedExecutorBuilder {
      * instances.</p>
      *
      * @return new instance of <code>ManagedExecutor</code>.
-     * @throws IllegalArgumentException if a thread context type that is
-     *         configured to be {@link #cleared} or {@link #propagated} is
-     *         unavailable. Also raises IllegalArgumentException if one or
-     *         more of the same context types appear in both the
-     *         {@link #cleared} set and the {@link #propagated} set.
-     * @throws IllegalStateException if the direct or indirect
+     * @throws IllegalStateException for any of the following error conditions
+     *         <ul>
+     *         <li>if one or more of the same context types appear in both the
+     *         {@link #cleared} set and the {@link #propagated} set</li>
+     *         <li>if a thread context type that is configured to be
+     *         {@link #cleared} or {@link #propagated} is unavailable</li>
+     *         <li>if the direct or indirect
      *         {@link org.eclipse.microprofile.concurrent.spi.ThreadContextProvider#getPrerequisites prerequisites}
-     *         of a <code>ThreadContextProvider</code> are unsatisfied,
-     *         or a provider has itself as a direct or indirect prerequisite,
-     *         or if more than one provider provides the same thread context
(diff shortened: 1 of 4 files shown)


njr-11:
@FroMage or @manovotn please review

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
```diff
@@ -44,9 +44,11 @@
      * instances.</p>
      *
      * @return new instance of <code>ManagedExecutor</code>.
-     * @throws IllegalArgumentException if the same thread context type is
-     *         present in, or inferred by, both the {@link #cleared} set
-     *         and the {@link #propagated} set.
+     * @throws IllegalArgumentException if a thread context type that is
```
manovotn: I think in this case having the `DefinitionException` would be more appropriate. There are no actual arguments in the method itself, hence `IllegalArgumentException` seems a bit off to me.

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
```diff
@@ -64,9 +64,10 @@
      * inclusion of the prerequisites, even if not explicitly specified.</p>
      *
      * <p>A <code>ManagedExecutor</code> must fail to inject, raising
-     * <code>DefinitionException</code> on application startup, if the same
-     * context type is implicitly or explicitly included in this set
-     * as well as in the set specified by {@link #propagated}.</p>
+     * <code>DefinitionException</code> on application startup,
```
manovotn: By `DefinitionException` I suppose you mean [the one from CDI](https://github.com/cdi-spec/cdi/blob/master/api/src/main/java/javax/enterprise/inject/spi/DefinitionException.java), right? If so, might be worth using a FQCN but I know this is just nitpicking really :)

njr-11: Good idea, it's best to be precise here.  I added a commit that that includes the fully qualified class name both here as well as for all mentions of DeploymentException where we were also lacking the package name.

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorBuilder.java
njr-11: We cannot use DefinitionException here because it would cause the builders to have a dependency on CDI and the whole reason the builders were added to the spec was to provide a way to use managed executors/thread context without CDI.  This was requested by Stéphane under https://github.com/eclipse/microprofile-sandbox/issues/18

That said, I'm certain open to a better exception class when the build method detects that an invalid combination of config settings were specified on the builder instance.  I'm sure this isn't the first time a spec with builders has encountered a pattern like this so maybe we can see if there is a precedent for it.

njr-11: After a bit of searching I found the following guidance in [Effective Java](https://books.google.com/books?id=ka2VUBqHiWkC&pg=PA15&lpg=PA15#v=onepage&q&f=false) written by Joshua Bloch: "_If any invariants are violated, the build method should throw an IllegalStateException. The exception's detail message should indicate which invariant is violated._"

I do also see where IllegalArgumentException is used in some existing specs, for example, in the [JsonbBuilder.build](https://static.javadoc.io/javax.json.bind/javax.json.bind-api/1.0/javax/json/bind/JsonbBuilder.html#build--)

It doesn't really concern me which exception type is used here so much as it does clearly defining what the error conditions are.  I can switch it over to IllegalStateException to be consistent with _Effective Java_.

njr-11:
@manovotn I added commits to address both review comments.  Please review again. Thanks!

njr-11:
@manovotn let me know if the added commits address the concerns in your review comments. If so, I'll go ahead and merge this

PR #38, opened 2018-11-06
Clean up ConcurrencyProvider and clearly indicate which parts are optional

### LINKED ISSUE #37: Clean up the ConcurrencyProvider SPI and make it clear which parts are optional
ConcurrencyProvider JavaDoc needs to make it clear to the implementer which parts are optional, or only needed under certain conditions.  There is also a bit of duplication between ConcurrencyProvider and ConcurrencyManager.  It should be possible to simplify this interface by cutting out some of the duplication and providing a few default implementations for existing methods, as well as adding some clarifying comments to JavaDoc, which will hopefully make it easier for provider implementers to understand.

njr-11:
pull fixes #37 

Removed newManagedExecutorBuilder/newThreadContextBuilder methods from ConcurrencyProvider because they duplicate the same methods on ConcurrencyManager.

Provided default implementation of getConcurrencyManager() which does what the JavaDoc already says it does.

Add clarifying JavaDoc to ConcurrencyProvider indicating what is optional and defaulting these methods to raising UnsupportedOperationException.

As as result of all of this, it is now possible to implement ConcurrencyProvider by implementing a single method:
getConcurrencyManager(ClassLoader)
And optionally, to implement the following methods:
getConcurrencyManagerBuilder()
register/releaseConcurrencyManager(ConcurrencyManager)

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
@@ -81,7 +81,7 @@ public interface ManagedExecutor extends ExecutorService {
      * @return a new {@link Builder} instance.
      */
     public static Builder builder() {
-        return ConcurrencyProvider.instance().newManagedExecutorBuilder();
+        return ConcurrencyProvider.instance().getConcurrencyManager().newManagedExecutorBuilder();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


njr-11:
Added a commit that moves the IllegalStateException for duplicate thread context provider types to getConcurrencyManager, which is where the providers are loaded per classloader.  It does not make sense to defer the reporting of this error to build() when it is already known that there is an error when loading the ConcurrencyManager.  Reporting the error closer to where the problem happens will make it more intuitive for the user to debug problems.

FroMage:
LGTM

PR #39, opened 2018-11-12
Use testng instead of junit to be consistent with other MP projects

aguibert:
Since most other microprofile projects use TestNG instead of JUnit for their TCK tests, I think it would be best if MP Concurrency also uses TestNG.

List of MP projects using TestNG:
- https://github.com/eclipse/microprofile-rest-client/tree/master/tck/pom.xml
- https://github.com/eclipse/microprofile-config/blob/master/tck/pom.xml
- https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/pom.xml
- https://github.com/eclipse/microprofile-health/blob/master/tck/pom.xml
- https://github.com/eclipse/microprofile-open-api/blob/master/tck/pom.xml

The only MP project I could find not using TestNG was MP Metrics:
- https://github.com/eclipse/microprofile-metrics/tree/master/tck


genie-microprofile:

njr-11:
@FrogMage this change to conform with the test framework used by the MicroProfile projects looks fine to me, but I want to be sure that you have a chance to review it given that you originally added the TCK as junit-based.

njr-11:
Discussed this on the call and decided to merge it.  If anyone has further comments, please open a new issue.

PR #50, opened 2018-11-28
create a spec doc starting from the current readme

njr-11:
Convert the existing readme, which we had been using as our spec documentation into official spec doc with separate pages.

Most of the readme, especially regarding application usage has been converted into the overview.asciidoc file.  The contextproviders.asciidoc file contains the parts of the readme that concern the SPI, to which an oversimplified example context provider is added.
The following pages are entirely new:
builders.asciidoc, examples.asciidoc, concurrencyprovider.asciidoc

TODO comments are intentionally included on the cdi page and the section of the overview for CDI injection because of the ongoing discussion that needs to be resolved before these can be written.  A separate pull request will be made for this work.

### DIFF
--- a/spec/pom.xml
+++ b/spec/pom.xml
@@ -88,7 +88,7 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <sourceDocumentName>microprofile-rest-client.asciidoc</sourceDocumentName>
+                    <sourceDocumentName>microprofile-concurrency.asciidoc</sourceDocumentName>
                     <sourceHighlighter>coderay</sourceHighlighter>
                     <attributes>
                         <license>Apache License v2.0</license>
--- a/spec/pom.xml
+++ b/spec/pom.xml
@@ -1,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright (c) 2018 Contributors to the Eclipse Foundation
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~    http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.eclipse.microprofile.concurrency</groupId>
+        <artifactId>microprofile-concurrency-parent</artifactId>
+        <version>1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>microprofile-concurrency-spec</artifactId>
+    <packaging>pom</packaging>
+
+    <name>MicroProfile Concurrency Specification</name>
+    <description>MicroProfile Concurrency :: Specification</description>
+
+    <properties>
+        <asciidoctor-maven.version>1.5.5</asciidoctor-maven.version>
+        <asciidoctorj-pdf.version>1.5.0-alpha.15</asciidoctorj-pdf.version>
+        <license>Apache License v 2.0</license>
+        <maven.build.timestamp.format>MMMM dd, yyyy</maven.build.timestamp.format>
+        <revisiondate>${maven.build.timestamp}</revisiondate>
+        <revremark>Draft</revremark>
+    </properties>
+
+    <build>
+        <defaultGoal>clean package</defaultGoal>
+        <plugins>
+            <plugin>
+                <groupId>org.asciidoctor</groupId>
+                <artifactId>asciidoctor-maven-plugin</artifactId>
+                <version>${asciidoctor-maven.version}</version>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.asciidoctor</groupId>
+                        <artifactId>asciidoctorj-pdf</artifactId>
+                        <version>${asciidoctorj-pdf.version}</version>
+                    </dependency>
+                </dependencies>
+                <executions>
+                    <execution>
+                        <id>generate-pdf-doc</id>
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>process-asciidoc</goal>
+                        </goals>
+                        <configuration>
+                            <backend>pdf</backend>
+                            <attributes>
+                                <revnumber>${project.version}</revnumber>
+                                <revremark>${revremark}</revremark>
+                                <revdate>${revisiondate}</revdate>
+                            </attributes>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>output-html</id>
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>process-asciidoc</goal>
+                        </goals>
+                        <configuration>
+                            <backend>html5</backend>
+                            <attributes>
+                                <revnumber>${project.version}</revnumber>
+                                <revremark>${revremark}</revremark>
+                                <revdate>${revisiondate}</revdate>
+                            </attributes>
+                        </configuration>
+                    </execution>
+                </executions>
+                <configuration>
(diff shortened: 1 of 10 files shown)


spec/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+                            <attributes>
+                                <revnumber>${project.version}</revnumber>
+                                <revremark>${revremark}</revremark>
+                                <revdate>${revisiondate}</revdate>
+                            </attributes>
+                        </configuration>
+                    </execution>
+                </executions>
+                <configuration>
+                    <sourceDocumentName>microprofile-rest-client.asciidoc</sourceDocumentName>
```
aguibert: this should be `microprofile-concurrency.asciidoc` insead of MP Rest client

njr-11: Thanks for catching this. I suppose that makes it obvious which project I copied this from when starting out.  I'll add a commit with the fix.

spec/src/main/asciidoc/contextproviders.asciidoc
```diff
@@ -0,0 +1,129 @@
...
+== Thread Context Providers
+
+The initial release of EE Concurrency assumed a single monolithic implementation of the full set of EE specifications that could thus rely on vendor-specific internals to achieve context propagation. However, in practice, open source implementations of various specs are often pieced together into a comprehensive solution.
+
+The thread context provider SPI is defined to bridge the gap, allowing any provider of thread context to publish and make available the type of thread context it supports, following a standard and predictable pattern that can be relied upon by a MicroProfile Concurrency implementation, enabling it to facilitate the inclusion of any generic thread context alongside the spec-defined thread context types that it captures and propagates.
+
+With this model, the provider of thread context implements the org.eclipse.microprofile.concurrent.spi.ThreadContextProvider interface and packages it in a way that makes it available to the ServiceLoader. ThreadContextProvider identifies the thread context type and provides a way to capture snapshots of thread context as well as for applying empty/cleared context to threads.
```
manovotn: Class names should have code formatting (using ``).
org.eclipse.microprofile.concurrent.spi.ThreadContextProvider -> `org.eclipse.microprofile.concurrent.spi.ThreadContextProvider`

I saw this multiple time throughout this PR, not sure if we want to do the same for class mentions without FQCN and for methods. IMO we should as it is then clear the spec refers to some bit of the actual code.

spec/src/main/asciidoc/contextproviders.asciidoc
```diff
@@ -0,0 +1,129 @@
...
+
+    public ThreadContextSnapshot clearedContext(Map<String, String> props) {
+        return new ThreadPrioritySnapshot(Thread.NORM_PRIORITY);
+    }
+}
+----
+
+==== ThreadContextSnapshot
+
+The interface, org.eclipse.microprofile.concurrent.spi.ThreadContextSnapshot, represents an immutable snapshot of thread context. The MicroProfile Concurrency implementation can request the context represented by this snapshot to be applied to any number of threads by invoking the the begin method. The begin method must save the previous context of the thread so that it can return an instance of org.eclipse.microprofile.concurrent.spi.ThreadContextController for one-time use by the MicroProfile Concurrency implementation to restore the previous context after the context represented by the snapshot is no longer needed on the thread.
```
manovotn: typo:
by invoking the the begin method -> by invoking the begin method

spec/src/main/asciidoc/examples.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+[[examples]]
+== MicroProfile Concurrency Examples
+
+This page includes some additional examples of spec usage.
```
manovotn: I suppose all these "pages" are put together into one big document, hence maybe using `This section` or `This part` would be more spot on.

spec/src/main/asciidoc/contextproviders.asciidoc
manovotn: `represents an immutable snapshot of thread context`

Not sure I understand this correctly, does it mean you assume you can use the same context repeatedly in several threads and every time you do that, all the threads will have the same initial state? E.g. that the actual context doesn't change?
This won't really be true for CDI and I am pretty sure transactions will be the same. Take transactions for starters, if you have a running tx you want to propagate, you capture the context and offload the execution elsewhere. However, once done this is done, tx is no longer running. How do you aim to propagate the same context repeatedly in another thread ("some time later")?
In CDI, we can propagate beans as in take what is in current context (talking CDI contexts now) and feed those instances into new thread. But those are ultimately the same beans so you can change their state which will be reflected in any further usage of that snapshot.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -0,0 +1,90 @@
...
+
+Applications should shut down ManagedExecutors that they build after they are no longer needed. The shutdown request serves as a signal notifying the container that resources can be safely cleaned up.
+
+=== Injection
+
+Instances of ManagedExecutor and ThreadContext can be injected into CDI beans via the `@Inject` annotation, for example:
+
+[source, java]
+----
+    @Inject ManagedExecutor executor;
```
manovotn: Have we actually specified how this "default" instance looks like?
E.g. when no config is present whatsoever.

spec/src/main/asciidoc/builders.asciidoc
```diff
@@ -0,0 +1,79 @@
...
+    // than a managed executor, context propagation is unpredictable,
+    // except for the contextFn action that we pre-contextualized using
+    // MicroProfile Concurrency's ThreadContext above.
+    stage = CompletableFuture.supplyAsync(supplier)
+                             .thenApplyAsync(function1)
+                             .thenApply(contextFn)
+                             ...
+----
+
+Builder instances retain their configuration after the build method is
```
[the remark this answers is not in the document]
njr-11: Good point, I didn't mean for this to be included as part of ThreadContext only.  I added a heading before this section for Reusing Builders and made the updates you suggested.

spec/src/main/asciidoc/builders.asciidoc
```diff
@@ -0,0 +1,79 @@
...
+----
+    threadContext = ThreadContext.builder()
+                        .propagated(ThreadContext.APPLICATION, ThreadContext.SECURITY)
+                        .cleared(ThreadContext.ALL_REMAINING)
+                        .build();
+
+    ...
+
+    Function<Long, Long> contextFn = threadContext.contextualFunction(x -> {
+        ... operations that requires security & application context
```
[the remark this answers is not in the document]
njr-11: fixed the grammar as suggested

spec/src/main/asciidoc/builders.asciidoc
```diff
@@ -0,0 +1,79 @@
...
+[[builders]]
+== Builders for ManagedExecutor and ThreadContext
+
+MicroProfile Concurrency provides a fluent builder API to programmatically obtain instances of ManagedExecutor and ThreadContext. Builder instances are obtained via static builder() methods on ManagedExecutor and ThreadContext.
+
+=== Example ManagedExecutor Builder Usage
+
+[source, java]
+----
+    public void init(ServletConfig config) {
```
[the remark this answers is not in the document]
njr-11: Yes, it definitely makes sense in this class where we are trying to illustrate the importance of using shutdown to end the managed executor's life cycle.  I'll update it.

spec/src/main/asciidoc/contextproviders.asciidoc
njr-11: Agree, I'll get the class/method references updated.

spec/src/main/asciidoc/contextproviders.asciidoc
njr-11: fixing

spec/src/main/asciidoc/examples.asciidoc
njr-11: Updating to "This section"

spec/src/main/asciidoc/contextproviders.asciidoc
njr-11: No, the intent when describing it as immutable was just to discourage the storing of other state (specifically, information about the previous context to restore) within a ThreadContextSnapshot instance because that would prevent the proper application/restore of context when reused for multiple threads.  Another way to think of it is that ThreadContextSnapshot is a pointer to thread context, for example, a transaction id. Even though the transaction might itself be in progress at one time, and at a later time has been committed or rolled back, the transaction id remains the same.
I'll work on coming up with a different wording for this, because we don't want to cause any confusion to context providers.

spec/src/main/asciidoc/overview.asciidoc
njr-11: Good point, we definitely need to do that (and the same for the default instance of ThreadContext).  For now, I'll document it as being equivalent to using the builder with all of the defaults, ManagedExecutor.builder().build()

njr-11:
@manovotn Thanks for the review comments. I added a commit to address them.  Let me know if any further changes are needed.

PR #51, opened 2018-11-29
scope and sharing for ManagedExecutorConfig and ThreadContextConfig

### LINKED ISSUE #46: Sharing of ManagedExecutor instances configured via CDI is non-intuitive
The [`@ManagedExecutorConfig` javadoc](https://github.com/eclipse/microprofile-concurrency/blob/master/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java#L31) currently states:

>  <p>Provides configuration for an injected {@link ManagedExecutor} instance
>  which is {@link javax.enterprise.context.Dependent dependent} scoped.
>  When an application has multiple injection points for {@link ManagedExecutor}
>  with matching configuration, the container injects the same instance.
>  For the purposes of matching, array attributes of this annotation are
>  considered as unordered sets where duplicate elements are ignored.</p>

I think this behavior should be removed (at least by default), because unless a user reads the config annotation javadoc (which is otherwise mostly self-explanatory), they may share ManagedExecutors in their application where they were not intending to do so.

Consider the following scenario:
```java
public class PaymentProcessor {
    @Inject
    @ManagedExecutorConfig(maxAsync = 5)
    ManagedExecutor paymentExectuor;
}

public class NotificationProcessor {
    @Inject
    @ManagedExecutorConfig(maxAsync = 5)
    ManagedExecutor notificationExecutor;
}
```

Before I closely read the spec, I would think that my payment and notification operations **each**  have maxAsync=5. However, since these annotations have equal values, the same underlying `ManagedExecutor` instance would be used, meaning that payment and notification operations have a **combined** maxAsync=5.

I would like to suggest that by default, `ManagedExecutor` instances injected via the config annotation are **not** shared. 

Since I am proposing to remove the ability to achieve the sharing use case, we need to consider what the alternative approaches are if we were to not share instances by default.  Here are the alternative options I can think of:

1. Use builder approach (no CDI):

```java
public class MyBean {
  public static final ManagedExecutor exec = ManagedExecutor.builder().maxAsync(5).build();
}

public class OtherBean {
  public void doSomething() {
    MyBean.exec.submit(...);
  }
}
```

2. Use a getter method:

```java
@ApplicationScoped
public class MyBean {
  @Inject
  @ManagedExecutorConfig(maxAsync = 5)
  ManagedExecutor exec;

  public ManagedExecutor getExec() {
    return exec;
  }
}

public class OtherBean {
  @Inject
  MyBean bean;

  public void doSomething() {
    bean.getExec().submit(...);
  }
}
```

3. Use a CDI producer+qualifier:

```java
@Qualifier
@Retention(RUNTIME)
@Target({TYPE, METHOD, FIELD, PARAMETER})
public @interface MyExec {}

@ApplicationScoped
public class MyBean {
  @Produces  @MyExec
  public ManagedExecutor getExec(@Inject @ManagedExecutorConfig(maxAsync = 5)
  ManagedExecutor exec) {
    return exec;
  }
}

public class OtherBean {
  @Inject
  @MyExec
  ManagedExecutor exec;

  public void doSomething() {
    exec.submit(...);
  }
}
```

I know this does make the ME sharing scenario more verbose, but I think this is acceptable because:
1. I don't see sharing a ManagedExecutor outside of a bean scope as a common use pattern
2. It eliminates the original scenario in this post where two injection points happened to have the same config and shared the same underlying `ManagedExecutor` and its constraints

njr-11:
Pull fixes #46 with agreed upon solution, which is briefly summarized as: inject the Config annotation in one place, and use name qualifier to share with other injection points.
See the discussion in the issue for more detail.


njr-11:
@manovotn could you review?

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextConfig.java
```diff
@@ -28,25 +28,42 @@
...
 
 /**
- * <p>Provides configuration for an injected {@link ThreadContext} instance,
- * which is {@link javax.enterprise.context.Dependent dependent} scoped.
- * When an application has multiple injection points for {@link ThreadContext}
- * with matching configuration, the container injects the same instance.
- * For the purposes of matching, array attributes of this annotation are
- * considered as unordered sets where duplicate elements are ignored.</p>
+ * <p>Annotates a CDI injection point for a {@link ThreadContext} such that the container
+ * creates a new instance, which is scoped within an application to its unique name.
```
manovotn: Same note as above about `scoped`

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextConfig.java
```diff
@@ -28,25 +28,42 @@
...
+ * The unique name is generated as the fully qualified class and field name of the
```
manovotn: Same note as above, separator?

api/src/main/java/org/eclipse/microprofile/concurrent/NamedContext.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * void doSomething(&commat;Inject &commat;NamedContext("secContext") ThreadContext secContextPropagator) {
+ *     ...
+ * }
+ * </code></pre>
+ */
+@Qualifier
+@Retention(RUNTIME)
+@Target({ FIELD, METHOD, PARAMETER, TYPE })
+public @interface NamedContext {
```
manovotn: Do we need two different qualifiers? IMO it's redundant if we can think of universal name (in CDi typesafe resolution takes a type and a qualifier, hence it should cause no conflicts in that regard). Anything that indicates that it has a name (though *not* just `@Named`, that's overloaded).
First things I can think of are `@UniqueName`, `@NameIdentifier`, `@SharedInstance`...probably not good candidates but about as far as I can get without finishing my coffee ;)

njr-11: I thought about this same thing myself before making the changes, and `@Named` seemed like an excellent choice until I realized it was already taken.  My other thoughts were that generic names for this sort of pattern seemed more the domain of the CDI spec rather than something that ought to be defined by a single specification using CDI, and so I opted for qualifiers that are concurrency-specific, but I couldn't think of single name that fit both.  If we do go with a generic name, another option is `@NamedInstance` combining two of your suggestions.

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
```diff
@@ -28,22 +28,47 @@
...
 /**
- * <p>Provides configuration for an injected {@link ManagedExecutor} instance
- * which is {@link javax.enterprise.context.Dependent dependent} scoped.
- * When an application has multiple injection points for {@link ManagedExecutor}
- * with matching configuration, the container injects the same instance.
- * For the purposes of matching, array attributes of this annotation are
- * considered as unordered sets where duplicate elements are ignored.</p>
+ * <p>Annotates a CDI injection point for a {@link ManagedExecutor} such that the container
+ * creates a new instance, which is scoped within an application to its unique name.
+ * The unique name is generated as the fully qualified class and field name of the
```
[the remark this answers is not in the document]
njr-11: Good catch, we definitely need to document the separator.  I did a quick search on the separator used by MP Config to refer to fields and it appears to be `.`, so I'll document it as that.

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
```diff
@@ -28,22 +28,47 @@
...
 
 /**
- * <p>Provides configuration for an injected {@link ManagedExecutor} instance
- * which is {@link javax.enterprise.context.Dependent dependent} scoped.
- * When an application has multiple injection points for {@link ManagedExecutor}
- * with matching configuration, the container injects the same instance.
- * For the purposes of matching, array attributes of this annotation are
- * considered as unordered sets where duplicate elements are ignored.</p>
+ * <p>Annotates a CDI injection point for a {@link ManagedExecutor} such that the container
+ * creates a new instance, which is scoped within an application to its unique name.
```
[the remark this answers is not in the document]
njr-11: This change is fine. I'll switch the wording

njr-11:
@manovotn I've added a commit that should address all of your review comments.

PR #53, opened 2018-12-04
add spec doc for CDI annotations

### LINKED ISSUE #43: Convert the content of our README.adoc into the start of an official spec document
We have been using our README.adoc file as a placeholder for the Concurrency spec doc.
Need to get this converted to official spec doc in a /spec/ folder as the other MicroProfile specs are doing.

njr-11:
Write spec documentation for the CDI annotations.
Also, while writing this, I noticed a few code examples in JavaDoc that are invalid and won't compile, so I'm fixing those as well. However, later I've moved those to pull #57 so that they can be independent of whether this version of the spec doc is used.

Pull fixes #43

### DIFF
--- a/spec/src/main/asciidoc/cdi.asciidoc
+++ b/spec/src/main/asciidoc/cdi.asciidoc
@@ -17,6 +17,75 @@
 [[concurrencycdi]]
 == CDI Injection
 
-If a CDI implementation is available, instances of `ManagedExecutor` and `ThreadContext` can be injected as CDI beans.
+If a CDI implementation is available, instances of `ManagedExecutor` and `ThreadContext` can be injected as CDI beans. 
 
-TODO write the documentation after discussion on Config annotations is resolved
+=== Injecting ManagedExecutor
+
+The application can annotate injection points of `ManagedExecutor` with both, either or neither of,
+
+- `@NamedInstance` - This annotation is a standard CDI qualifier. It has a single value, which is the unique name for a `ManagedExecutor` instance. All injection points within an application that specify the same name inject the same instance.
+
+- `@ManagedExecutorConfig` - This annotation supplies configuration for an instance of `ManagedExecutor` that is built and managed by the container. The same instance is shared within an application across injection points that have the same unique name. The unique name is determined by the value `@NamedInstance` annotation if present. Otherwise, it is generated from the fully qualified class name and field name of the injection point, with the `.` character as the delimiter.
+
+When an injection point for `ManagedExecutor` is annotated with only the `@NamedInstance` qualifier, the container injects an existing instance of `ManagedExecutor` that is shared across all injection points in the application that have the same name.
+
+When an injection point for `ManagedExecutor` is annotated with only the `@ManagedExecutorConfig` annotation, the container builds a single instance of `ManagedExecutor` matching the configuration attributes of the `@ManagedExecutorConfig` annotation. The container injects this same instance into the annotated injection point, managing its life cycle, and shutting it down when the application stops.
+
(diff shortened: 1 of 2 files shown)


njr-11:
@manovotn could you review these additions to the spec doc to include the CDI config annotations?  Note that under this pull I also fixed buggy examples the ManagedExecutorConfig/ThreadContextConfig JavaDoc.  When I went to copy from there to create some of the spec examples, I realized they won't compile that way, so corrected those as well.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -17,6 +17,75 @@
...
+
+When an injection point for `ManagedExecutor` is annotated with both `@ManagedExecutorConfig` and `@NamedInstance`, the container builds a single instance of `ManagedExecutor` matching the configuration attributes of the `@ManagedExecutorConfig` annotation. The container injects this same instance into the annotated injection point, in addition to all other injection points that are annotated with `@NamedInstance` that specify the same name. The container manages the life cycle of the `ManagedExecutor` instance, shutting it down when the application stops.
+
+When an injection point for `ManagedExecutor` is annotated with neither of the above, the container injects a default instance of `ManagedExecutor` which is shared within the application across all similar injection points (annotated with neither of the above). The configuration of this instance matches the default attribute values that are used by `@ManagedExecutorConfig()`.
```
manovotn: `...which is shared within the application across all similar injection points...`
This goes against the default sharing you and @aguibert argued against. It is also contradictory to only sharing instances with same name on them and gives you no way to request several different default executors in your application.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,11 +79,60 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
...
+    @ManagedExecutorConfig(propagated = { ThreadContext.SECURITY, ThreadContext.APPLICATION })
+    ManagedExecutor executor1;
+    ...
+    @Inject
+    void setCompletableFuture(@NamedInstance("exec1") ManagedExecutor exec) {
+        completableFuture = exec.newIncompleteFuture();
+    }
+----
+
+=== Injection of Custom Built Instances
```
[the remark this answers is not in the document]
njr-11: Although the producer approach is generally obvious to CDI users, I provided this example for a couple of reasons that are specific to ManagedExecutors.
First, I wanted to emphasize that if you build your own managed executor instance for a CDI producer, don't forget to shut it down.
Second, to encourage the use of ApplicationScoped rather than other scopes.
I don't believe the user should need to annotate this with `@Alternative` or `@Priority` at all for this example because it uses the `@NamedInstance` qualifier.  They only matching injection points for this producer will be the ones annotated with `@NamedInstance("exec2")` as shown in the example injection point corresponding to the producer.
This was intended to be a good example of how the user can properly take advantage of CDI producer in a way that fits naturally with the MP Concurrency spec.

spec/src/main/asciidoc/cdi.asciidoc
njr-11: I'm fine with removing the language about sharing here.  I was thinking it might be more efficient given that the default configuration for ManagedExecutor has unlimited maxAsync/maxQueued (unconstrained) and so which instance is used wouldn't make much difference in that case.  However, if MP Config is used to override, then there could be concurrency constraints configured and we would need separate instances after all.

njr-11:
It doesn't sound like we are going to use this version of the doc, but I went ahead and updated commit so that it would at least be consistent with the review comments.  Also, change to correct the errors in the Config JavaDoc where the examples don't compile was moved to #57 in order to ensure that bug is corrected regardless.

njr-11:
Closing out as duplicate, given that this documentation was already provided by Matej under #54

Issue #31, opened 2018-10-19
External configuration of what default context propagation means


manovotn:
Based on discussions around issue #27, e.g. what contexts should be propagated by default, there was a discussion about the proc/cons of having this "baked" into spec versus leaving it up to implementation.

While both sides had good arguments, the way to go could be to hard-code default into spec (whatever the default will be) and allow for external configuration via some file - yaml, xml - which would be able to override the defaults. Basically implementations would look for this will and follow what's in it, if there is no setup, spec defaults will be used.

During the mtg, someone also mentioned that we could use MP Config - but I am not really sure what was exactly meant by that in practice, feel free to expand on that bit.

FroMage:
I suggested that. I seem to recall that all MP specs have to use MP Config for configuration, instead of doing their own thing.

tomjenkinson:
Just to respond to *could be to hard-code default into spec"
+1, the default should be in the spec

aguibert:
first off, I agree that the spec should define a default set of contexts.

As for configuration, MP Config is mainly useful when you want to control configuration that varies based on where the application is deployed (i.e. dev/test/staging/prod).  If configuration remains constant through all environments, annotation-based configuration is superior because it is type-safe and enforced by the compiler.
For example, it doesn't make sense to only want Security context in prod but not dev environments.

However, for some other configuration attributes in this spec, such as `@ManagedExecutorConfig(maxAsync=5)`, I expect it will be common to want to reconfigure these values based on the environment. 
Ultimately, I think it would be good to allow configuration through MP Config for **all** default instance values for consistency.  So something like:
```
org.eclipse.microprofile.concurrency.DefaultThreadContext.value=SECURITY,TRANSACTION
org.eclipse.microprofile.concurrency.DefaultManagedExecutor.maxAsync=5
```

The pattern being:
`org.eclipse.microprofile.concurrency.Default<IFC_NAME>.<ATTR_NAME>=<VALUE>`

njr-11:
Under #85, I wrote up the spec doc for integration with MP Config based on what we decided on the prior MP Concurrency call.  I tried to cover all of the scenarios I could think of.  @manovotn could you review?  I will also locate an MP Config expert to review as well, to be certain that we are following conventions appropriately.

PR #54, opened 2018-12-05
Issue #43 Spec doc for CDI injection.

manovotn:
Two things to note here are that default configured instances are not shared (which IMO follows the result of our previous discussions) unless you name them and that I deliberately left out any mention of custom producers as the possibility of that is simply implied by using CDI as a mechanism.

@njr-11 here is my take on the CDI part. I had it mostly completed before but wanted to do some polishing before sending it. Please let me know what you think about it.

### DIFF
--- a/spec/src/main/asciidoc/cdi.asciidoc
+++ b/spec/src/main/asciidoc/cdi.asciidoc
@@ -18,5 +18,160 @@
 == CDI Injection
 
 If a CDI implementation is available, instances of `ManagedExecutor` and `ThreadContext` can be injected as CDI beans.
+Injection points with the `@NamedInstance` qualifier have a name that is explicitly declared by the qualifier.
+Injection points that have no CDI qualifier on them have a unique name which is deduced from the injection point itself.
+Container provides, per unique name, an instance of either `ManagedExecutor` and/or `ThreadContext` unless the application specifies a CDI producer with `@NamedInstance` qualifier, given unique name and respective type.
 
-TODO write the documentation after discussion on Config annotations is resolved
+=== Injecting `ManagedExecutor`
+
+The simplest use case is a new executor instance per injection point, where the configuration is equivalent to the default values of the `@ManagedExecutorConfig` annotation.
+
+[source, java]
+----
+@Inject
+ManagedExecutor executor;
+----
+
+==== Configuring `ManagedExecutor`
+
+Oftentimes there is a need to configure the executor instance to limit the number of asynchronous tasks that it can start or define which contexts to propagate.
+The `@ManagedExecutorConfig` annotation allows you to configure all aspects of an injected `ManagedExecutor` instance.
+Note that when using this annotation only, new instance of `ManagedExecutor` is created for every injection point even if those injection points have equivalent configuration.
+
+[source, java]
+----
+@Inject
+@ManagedExecutorConfig(maxAsync = 5, propagated = ThreadContext.SECURITY)
+ManagedExecutor configuredExecutor;
+----
+
+==== Sharing `ManagedExecutor` Instances
+
+The CDI qualifier, `@NamedInstance`, enables you to share the same instance of `ManagedExecutor` across multiple injection points.
+The `@NamedInstance` qualifier has a single parameter which specifies the unique name that is assigned to the given executor configuration.
+Having one injection point annotated with both `ManagedExecutorConfig` and `@NamedInstance`, all other injection points within the application that have a matching `@NamedInstance` qualifier will have the same `ManagedExecutor` instance injected.
+Assuming two beans wish to share an executor instance, the code looks as follows:
+
+[source, java]
+----
+@ApplicationScoped
+public class MyFirstBean {
+
+  @Inject
+  @NamedInstance("myExecutor")
+  @ManagedExecutorConfig(maxAsync = 5, propagated = ThreadContext.SECURITY)
+  ManagedExecutor executor;
+
+}
+
+@ApplicationScoped
+public class MySecondBean {
+
+  @Inject
+  @NamedInstance("myExecutor")
+  ManagedExecutor sharedExecutor;
+
+}
+----
+
+Since the configuration of `ManagedExecutor` is optional, a shared `ManagedExecutor` with default configuration is also a possibility.
+
+[source, java]
+----
+@ApplicationScoped
+public class FooBean {
+
+  @Inject
+  @NamedInstance("sharedDefaultME")
+  ManagedExecutor executor;
+
+}
+
+@ApplicationScoped
+public class BarBean {
+
+  @Inject
+  @NamedInstance("sharedDefaultME")
+  ManagedExecutor sharedExecutor;
+
+}
+----
+
+If multiple configurations of `ManagedExecutor` are assigned the same name, the container detects it and raises a `javax.enterprise.inject.spi.DefinitionException`.
+
+=== Injecting `ThreadContext`
+
+The simplest use case is a new `ThreadContext` instance per injection point, where the configuration is equivalent to the default values of the `@ThreadContextConfig` annotation.
+
+[source, java]
+----
+@Inject
+ThreadContext context;
+----
+
+==== Configuring `ThreadContext`
+
+Oftentimes there is a need to configure the `ThreadContext` instance to specify which contexts to propagate.
+The `@ThreadContextConfig` annotation allows you to configure all aspects of an injected `ThreadContext` instance.
+Note that when using this annotation only, new instance of `ThreadContext` is created for every injection point even if those injection points have equivalent configuration.
+
+[source, java]
+----
+@Inject
+@ThreadContextConfig(propagated = ThreadContext.SECURITY)
+ThreadContext configuredContext;
+----
+
+==== Sharing `ThreadContext` Instances
+
+The CDI qualifier, `@NamedInstance`, enables you to share the same instance of `ThreadContext` across multiple injection points.
+The `@NamedInstance` qualifier has a single parameter which specifies the unique name that is assigned to the given executor configuration.
+Having one injection point annotated with both `ThreadContextConfig` and `@NamedInstance`, all other injection points within the application that have a matching `@NamedInstance` qualifier will have the same `ThreadContext` instance injected.
+Assuming two beans wish to share an executor instance, the code looks as follows:
+
+[source, java]
+----
+@ApplicationScoped
+public class MyFirstBean {
+
+  @Inject
+  @NamedInstance("myThreadContext")
+  @ThreadContextConfig(propagated = ThreadContext.SECURITY)
+  ThreadContext context;
+
+}
+
+@ApplicationScoped
+public class MySecondBean {
+
+  @Inject
+  @NamedInstance("myThreadContext")
+  ManagedExecutor sharedContext;
+
+}
+----
+
+Since the configuration of `ThreadContext` is optional, a shared `ThreadContext` with default configuration is also a possibility.
+
+[source, java]
+----
+@ApplicationScoped
+public class FooBean {
+
+  @Inject
+  @NamedInstance("sharedDefaultTC")
+  ThreadContext context;
+
+}
+
+@ApplicationScoped
+public class BarBean {
+
+  @Inject
+  @NamedInstance("sharedDefaultTC")
+  ThreadContext sharedContext;
+
+}
+----
+
+If multiple configurations of `ThreadContext` are assigned the same name, the container detects it and raises a `javax.enterprise.inject.spi.DefinitionException`.
--- a/spec/src/main/asciidoc/overview.asciidoc
+++ b/spec/src/main/asciidoc/overview.asciidoc
@@ -64,7 +64,11 @@ Applications should shut down instances of `ManagedExecutor` that they build aft
 
 === Injection
 
-Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI beans via the `@Inject` annotation, for example:
+Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI beans via the `@Inject` annotation.
+Injection points with the `@NamedInstance` qualifier have a name that is explicitly declared by the qualifier.
+Injection points that have no CDI qualifier on them have a unique name which is deduced from the injection point itself.
+Container provides, per unique name, an instance of either `ManagedExecutor` and/or `ThreadContext` unless the application specifies a CDI producer with `@NamedInstance` qualifier, given unique name and respective type.
+ for example:
 
 [source, java]
 ----
@@ -79,11 +83,50 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
     unmanagedCompletionStage.thenApply(threadContext.contextualFunction(function3));
 ----
 
-The container provides default instances of `ManagedExecutor` and `ThreadContext`, as injected above, which are shared within an application with all other injection points for `ManagedExecutor` and `ThreadContext` that do not specify a qualifier. Their configuration is equivalent to `ManagedExecutor.builder().build()` and `ThreadContext.builder().build()`.
+In the absence of other qualifiers and annotations, the container creates and injects a new default instance of `ManagedExecutor` or `ThreadContext` per injection point, as shown above.
+The configuration of these instances is equivalent to the default values of `@ManagedExecutorConfig` and `ThreadContextConfig`.
 
 === Injection of Configured Instances
 
-TODO write this section after discussion about the Config annotations is resolved.
+The `@ManagedExecutorConfig` and `@ThreadContextConfig` annotations provide the ability to configure the injection points with desired capabilities.
+With just the config annotation in place, every injection point is assigned a new instance.
+
+[source, java]
+----
+    @Inject 
+    @ManagedExecutorConfig(maxAsync = 5, propagated = ThreadContext.SECURITY)
+    ManagedExecutor configuredExecutor;
+    ...
+    CompletableFuture<Long> stage = configuredExecutor
+        .newIncompleteFuture()
+        .thenApply(function)
+        .thenAccept(consumer);
+    stage.completeAsync(supplier);
+----
+
+=== Sharing Configured Instances
+
+Configured instances of `ManagedExecutor` and `ThreadContext` are shared based on their unique name, which is assigned by the CDI qualifier, `@NamedInstance`.
+When one injection point is annotated with both a configuration and a `@NamedInstance` qualifier, the application can inject the same instance elsewhere by annotating other injection points with just the matching qualifier.
+In the example below, the `executor1` injection point defines a configuration and a name `myExec` for a configured `ManagerExecutor` instance.
+The other injection points, `executor2` and `executor3`, share the same `ManagedExecutor` instance by using the `@NamedInstance` qualifier with the name `myExec`.
+
+[source, java]
+----
+    @Inject
+    @NamedInstance("myExec")
+    @ManagedExecutorConfig(propagated = { ThreadContext.SECURITY, ThreadContext.APPLICATION })
+    ManagedExecutor executor1;
+    ... // in some other bean
+    @Inject
+    void setCompletableFuture(@NamedInstance("myExec") ManagedExecutor executor2) {
+        completableFuture = executor2.newIncompleteFuture();
+    }
+    ... // in yet another bean
+    @Inject
+    @NamedInstance("myExec")
+    ManagedExecutor executor3;
+----
 
 === Thread Context Provider SPI
 


genie-microprofile:

manovotn:
I'll look into the ECA compliance (and other things around that) first thing tomorrow.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+  ManagedExecutor sharedExecutor;
+
+}
+----
+
+If multiple configurations of `ManagedExecutor` are assigned the same name, the container detects it and raises a `javax.enterprise.inject.spi.DefinitionException`.
+
+=== Injecting `ThreadContext`
+
+Application can inject instances of `ThreadContext` using CDI with the option to share same `ThreadContext` between multiple injection points.
```
njr-11: Similar to comment on Injecting Manager, this sentence is confusing where currently placed (introducing an example of not sharing).  I'd suggest moving it to the section on "Sharing `ThreadContext` Instances", but you already covered this information there, so it seems best to just omit from the current section.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+The simplest use case will create a new `ThreadContext` instance for every such injection point and the configuration will be equal to default values of `@ThreadContextConfig` annotation.
```
njr-11: Minor grammar fixes, similar to the ManagedExecutor equivalent sentence,

The simplest use case is a new `ThreadContext` instance per injection point, where the configuration is equivalent to the default values of the @ThreadContextConfig annotation.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+
+[source, java]
+----
+@Inject
+ThreadContext context;
+----
+
+==== Configuring `ThreadContext`
+
+Oftentimes there is need to fine-grain the `ThreadContext` instance and modify what contexts are to be propagated.
```
njr-11: Oftentimes there is a need to configure the `ThreadContext` instance to specify which contexts to propagate.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+This is where `@ThreadContextConfig` annotation comes into play allowing to configure all the aspects of an injected `ThreadContext` instance.
```
njr-11: The `@ThreadContextConfig` annotation configures all aspects of an injected `ThreadContext` instance.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+Note that using just this annotation, new instance of `ThreadContext` is created for every injection point even should those injection points have equal configuration.
```
njr-11: When using this annotation without specifying the `@NamedInstance` qualifier, a new instance of `ThreadContext` is injected for every such injection point, even if those injection points have equivalent configuration.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+----
+@Inject
+@ThreadContextConfig(propagated = ThreadContext.SECURITY)
+ThreadContext configuredContext;
+----
+
+==== Sharing `ThreadContext` Instances
+
+Last but not least, sometimes it is desirable to share the same instance of `ThreadContext` among several injection points.
+For such use case, the specification provides a CDI qualifier `@NamedInstance` which takes one parameter - a unique name assigned to the given `ThreadContext` configuration.
```
njr-11: Recommend rewriting the first two sentences as:

The CDI qualifier, `@NamedInstance`, enables you to share the same instance of `ThreadContext` across multiple injection points. The `@NamedInstance` qualifier has a single parameter which specifies the unique name that is assigned to the given `ThreadContext` configuration.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+Having one injection point that defines both, the name using `@NamedInstance` and the configuration using `@ThreadContextConfig`, application can then inject the same instance repeatedly using the qualifier.
```
njr-11: When an injection point is annotated with both `@ThreadContextConfig` and `@NamedInstance`, the container creates a `ThreadContext` instance matching the configuration and injects this same instance into all injection points within the application that are annotated with a matching `@NamedInstance` qualifier.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+public class MySecondBean {
+
+  @Inject
+  @NamedInstance("myThreadContext")
+  ManagedExecutor sharedContext;
+
+}
+----
+
+Since the configuration of `ThreadContext` is optional, a shared `ThreadContext` with default configuration is also a possibility.
```
njr-11: Similar to the earlier pattern, this seems wrong. In the example that follows, the `@NamedInstance("sharedDefaultTC")` qualifier does not match any instance, so wouldn't this be an injection failure?

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,12 +79,52 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
     unmanagedCompletionStage.thenApply(threadContext.contextualFunction(function3));
 ----
 
-The container provides default instances of `ManagedExecutor` and `ThreadContext`, as injected above, which are shared within an application with all other injection points for `ManagedExecutor` and `ThreadContext` that do not specify a qualifier. Their configuration is equivalent to `ManagedExecutor.builder().build()` and `ThreadContext.builder().build()`.
+The container allows to inject default instances of `ManagedExecutor` and `ThreadContext` which do not leverave either `@NamedInstance` or `ManagedExecutorConfig`/`ThreadContextConfig` annotations, as shown above.
```
njr-11: In the absence of other qualifiers and annotations, the container creates and injects a new default instance of `ManagedExecutor` or `ThreadContext` per injection point, as shown above.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,12 +79,52 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
...
+These instances are not shared and container creates a new instance for each injection point of `ManagedExecutor` or `ThreadContext`.
```
njr-11: Usage of the term 'shared' here is slightly confusing.  It is shared with the same injection point (for example, if in a request scoped bean that is created/destroyed often), but not with other injection points.  That said, given the requested update to the previous sentence, the current sentence should now be redundant and can be omitted.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,12 +79,52 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
...
+The configuration of these instances is equal to default values present in `@ManagedExecutorConfig` and `ThreadContextConfig`.
```
njr-11: The configuration of these instances is equivalent to the default values of `@ManagedExecutorConfig` and `@ThreadContextConfig`.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,12 +79,52 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
...
 
 === Injection of Configured Instances
 
-TODO write this section after discussion about the Config annotations is resolved.
+The `@ManagedExecutorConfig` and `@ThreadContextConfig` provide the ability to configure the injection points with desired capabilities.
```
njr-11: The `@ManagedExecutorConfig` and `@ThreadContextConfig` annotations provide the ability to configure the injection points with desired capabilities.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,12 +79,52 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
...
+    CompletableFuture<Long> stage = configuredExecutor
+        .newIncompleteFuture()
+        .thenApply(function)
+        .thenAccept(consumer);
+    stage.completeAsync(supplier);
+----
+
+=== Sharing Configured Instances
+
+Configured instances of `ManagedExecutor` and `ThreadContext` can also be shared so long as they are given a CDI qualifier `@NamedInstance` which assigns them a unique name per configuration.
```
njr-11: Configured instances of `ManagedExecutor` and `ThreadContext` are shared by their unique name, which is assigned by the CDI qualifier, `@NamedInstance`.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,12 +79,52 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
...
+Given that one injection point contains a configuration and a name, application can inject the same instance elsewhere using just the CDI qualifier as an identifier.
```
njr-11: When one injection point is annotated with both a configuration and a `@NamedInstance` qualifier, the application can inject the same instance elsewhere by annotating other injection points with just the matching qualifier.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,12 +79,52 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
...
+In the example below, `executor1` injection point defines a configuration and a name `myExec` for a configured `ManagerExecutor` instance.
```
njr-11: In the example below, the `executor1` injection point defines a configuration and a name, `myExec`, for a configured `ManagerExecutor` instance.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,12 +79,52 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
...
+The other injection points, `executor2` and `executor3`, then get the same `ManagedExecutor` instance by using the `@NamedInstance` qualifier with the name `myExec`.
```
njr-11: The other injection points, `executor2` and `executor3`, share the same `ManagedExecutor` instance by using the `@NamedInstance` qualifier with the name `myExec`.

spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -79,12 +79,52 @@ Instances of `ManagedExecutor` and `ThreadContext` can be injected into CDI bean
...
+[source, java]
+----
+    @Inject
+    @NamedInstance("myExec")
+    @ManagedExecutorConfig(propagated = { ThreadContext.SECURITY, ThreadContext.APPLICATION })
+    ManagedExecutor executor1;
+    ... // in some other bean
+    @Inject
+    void setCompletableFuture(@NamedInstance("myExec") ManagedExecutor executor2) {
+        completableFuture = exec.newIncompleteFuture();
```
njr-11: completableFuture = executor2.newIncompleteFuture();

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
 
 If a CDI implementation is available, instances of `ManagedExecutor` and `ThreadContext` can be injected as CDI beans.
 
-TODO write the documentation after discussion on Config annotations is resolved
\ No newline at end of file
+=== Injecting `ManagedExecutor`
+
+Application can inject instances of `ManagedExecutor` using CDI with the option to share same executors between multiple injection points.
+The simplest use case will create a new executor instance for every such injection point and the configuration will be equal to default values of `@ManagedExecutorConfig` annotation.
```
aguibert: (This ties into one of the later comments I left about treating the default `@ManagedExecutorConfig` annotation differently than the absence of the annotation entirely, it would be helpful to read that comment first)

If we stick with the pattern that just `@Inject @NamedInstance` declares  _usage_, and `@Inject @NamedInstance @ManagedExecutorConfig` declares _usage and definition_, I think it would make more sense to have 1 shared `ManagedExecutor` across all default injection points (`@Inject ManagedExecutor exec;`).

Since the default ManagedExecutor instance is unbounded (infinite maxAsnyc and infinite maxQueued), it doesn't really matter if we share these instances or not, but conceptually it makes more sense to only get a dedicated instances when you are using `@ManagedExecutorConfig` (which is analogous to a Producer).  

What do others think on this?

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+
+Since the configuration of `ThreadContext` is optional, a shared `ThreadContext` with default configuration is also a possibility.
+
+[source, java]
+----
+@ApplicationScoped
+public class FooBean {
+
+  @Inject
+  @NamedInstance("sharedDefaultTC")
```
aguibert: same comment here as with ManagedExecutors.  Also, if we are going to show examples of invalid code in our spec doc, we should make it more obvious that the code example is bad, or point out via in-line comments what is wrong with the code example.  
Reading this, I'm not certain if the following text about a DefinitionException is describing this code example or not.

spec/src/main/asciidoc/overview.asciidoc
aguibert: I am in favor of having `@Default` injection points use a shared instance, rather than one instance per injection point.  Conceptually I think of `@NamedInstance` as qualifier, and `@XXConfig` being a "synthetic producer" for the associated qualifier (whether the qualifier is explicitly declared or inferred via FQCN+field).

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
 
 If a CDI implementation is available, instances of `ManagedExecutor` and `ThreadContext` can be injected as CDI beans.
 
-TODO write the documentation after discussion on Config annotations is resolved
\ No newline at end of file
+=== Injecting `ManagedExecutor`
+
+Application can inject instances of `ManagedExecutor` using CDI with the option to share same executors between multiple injection points.
```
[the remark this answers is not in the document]
manovotn: I originally wanted to give an overview of options, but we might as well drop it.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+----
+@Inject
+ManagedExecutor executor;
+----
+
+==== Configuring `ManagedExecutor`
+
+Oftentimes there is need to fine-grain the executor instance and change the number of maximum asynchronous tasks or define what contexts are to be propagated.
+This is where `@ManagedExecutorConfig` annotation comes into play allowing to configure all the aspects of an injected `ManagedExecutor` instance.
+Note that using just this annotation, new instance of `ManagedExecutor` is created for every injection point even should those injection points have equal configuration.
```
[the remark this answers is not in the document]
manovotn: I didn't want to get ahead of myself and mention @NamedInstance before actually explaining it. I would prefer to keep it that way. I'll reword the ending as suggested though.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+@Inject
+@ManagedExecutorConfig(maxAsync = 5, propagated = ThreadContext.SECURITY)
+ManagedExecutor configuredExecutor;
+----
+
+==== Sharing `ManagedExecutor` Instances
+
+Last but not least, sometimes it is desirable to share the same instance of `ManagedExecutor` among several injection points.
+For such use case, the specification provides a CDI qualifier `@NamedInstance` which takes one parameter - a unique name assigned to the given executor configuration.
+Having one injection point that defines both, the name using `@NamedInstance` and the configuration using `@ManagedExecutorConfig`, application can then inject the same instance repeatedly using the qualifier.
```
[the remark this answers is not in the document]
manovotn: I wanted to define this more from a user perspective, e.g. that you need to have one IP with name and config and the rest with just config. That the container provides shared bean is an "impl detail".
Nonetheless, will alter the wording to improve it.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,158 @@
...
+public class MySecondBean {
+
+  @Inject
+  @NamedInstance("myExecutor")
+  ManagedExecutor sharedExecutor;
+
+}
+----
+
+Since the configuration of `ManagedExecutor` is optional, a shared `ManagedExecutor` with default configuration is also a possibility.
```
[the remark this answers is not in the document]
manovotn: Not necessarily, we said that you can `@Inject ManagedExecutor` without config and it has default settings.
Furthermore we said (actually you fought for this to be default) that **sharing is not default behaviour**.

So what's the logic behind making default-configured instances shared without actually having a name when nothing else is shared?
Besides, assuming your proposal is in place, how do you aim to cover the use case when user wants default executor and wants it shared? You will have to do it manually (producer, builder) for no good reason.

The solution I have here follows the injection/configuration we have in place for all other cases. Ultimately _not having the config annotation is the same as having the annotation and placing the default values in it, right_?

spec/src/main/asciidoc/cdi.asciidoc
manovotn: > Reading this, I'm not certain if the following text about a DefinitionException is describing this code example or not.

All proposed text describes how to achieve a certain behaviour and then shows it with bits of code and this is no exception. Besides this code won't throw the exception because similar configuration is not given a different name.

spec/src/main/asciidoc/overview.asciidoc
manovotn: Agreed, that's a good reword you proposed

spec/src/main/asciidoc/overview.asciidoc
manovotn: `are shared by their unique` sounds a bit weird.
I'd go for `are shared based on their unique name, which is...`

spec/src/main/asciidoc/cdi.asciidoc
manovotn: > Since the default ManagedExecutor instance is unbounded 

That's what we designed now, but note that we discussed https://github.com/eclipse/microprofile-concurrency/issues/31 and external config of default settings. With that in place, defaults could differ (depending on what we allow there, but I see no reason not to permit changing all parameters).

>  I think it would make more sense to have 1 shared ManagedExecutor across all default injection points

Like I said (elsewhere on this PR), this approach is the opposite to what we drafted for sharing. The argument back there was that sharing by default is not intuitive for user. Hence sharing defaults is non-intuitive as well, especially given that in all other cases it behaves differently.

Impl-wise this isn't a problem. You can detect injection points that have a name but no config linked to them. In such case you simply provide default with a name attached to it to enable sharing.

manovotn:
@njr-11 updated the PR with you comments, some rewording I took literally, some I adjusted a bit (leaving a comment to mark those), give it a spin if you will.

Also signed the ECA and added the signoff to commits (hence the force push).

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,156 @@
...
+----
+@Inject
+ManagedExecutor executor;
+----
+
+==== Configuring `ManagedExecutor`
+
+Oftentimes there is a need to configure the executor instance to limit the number of asynchronous tasks that it can start or define which contexts to propagate.
+The `@ManagedExecutorConfig` annotation allows to configure all aspects of an injected `ManagedExecutor` instance.
+Note that using this annotation only, new instance of `ManagedExecutor` is created for every injection point even if those injection points have equivalent configuration.
```
njr-11: Minor grammar updates:
allows to configure --> allows you to configure
Note that using --> Note that when using
new instance of --> a new instance of

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,156 @@
...
+@Inject
+@ManagedExecutorConfig(maxAsync = 5, propagated = ThreadContext.SECURITY)
+ManagedExecutor configuredExecutor;
+----
+
+==== Sharing `ManagedExecutor` Instances
+
+The CDI qualifier, `@NamedInstance`, enables you to share the same instance of `ManagedExecutor` across multiple injection points.
+The `@NamedInstance` qualifier has a single parameter which specifies the unique name that is assigned to the given executor configuration.
+Having one injection point annotated with both `ManagedExecutorConfig` and `@NamedInstance`, all other injection points within the application that have a matching `@NamedInstance` qualifier will have the same `ManagedExecutor` instance injected.
```
njr-11: Grammar:
Having one injection point annotated --> When one injection point is annotated

spec/src/main/asciidoc/cdi.asciidoc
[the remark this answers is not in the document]
njr-11: This concern (which Andy also addresses in a subsequent comment further down) will be discussed using comments on the main issue.  I added a summary there with my understanding of the two proposed options.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -19,4 +19,156 @@
...
+----
+@Inject
+ThreadContext context;
+----
+
+==== Configuring `ThreadContext`
+
+Oftentimes there is a need to configure the `ThreadContext` instance to specify which contexts to propagate.
+The `@ThreadContextConfig` annotation allows to configure all aspects of an injected `ThreadContext` instance.
+Note that using this annotation only, new instance of `ThreadContext` is created for every injection point even if those injection points have equivalent configuration.
```
njr-11: Minor grammar updates:
allows to configure --> allows you to configure
Note that using --> Note that when using
new instance of --> a new instance of

spec/src/main/asciidoc/cdi.asciidoc
njr-11: Note: This falls under the same category as the similar section for ManagedExecutor that will be discussed under the main issue

manovotn:
@njr-11 Either my English decayed drastically over last few months, or vast majority of your comments are not grammatical mistakes but rather stylistic differences.

manovotn:
@njr-11, @aguibert I have updated the PR as we agreed on last meeting, both the `overview` and the `cdi` part now contain similar two sentences about container providing instances per unique name unless a producer was specified.

With this I hope we are good to proceed.

manovotn:
Ok, it is a minor change for affecting all IPs would look like a bug in the first place.
Updated the PR accordingly to your suggestion and squashed into one commit.

Issue #55, opened 2018-12-05
Introduce annotation literals for easy inline creation of annotations


manovotn:
I think we should introduce helper classes for users and integrators to be able to instantiate an annotation with given parameters programatically.

This is quite common in CDI, see for instance [`NamedLiteral` qualifier](https://github.com/cdi-spec/cdi/blob/master/api/src/main/java/javax/enterprise/inject/literal/NamedLiteral.java) (placed in separate class only because CDI doesn't "own" Named class, that javax.annotation)
or the [`@ApplicationScoped`](https://github.com/cdi-spec/cdi/blob/master/api/src/main/java/javax/enterprise/context/ApplicationScoped.java#L96) annotation which is not a qualifier. And many more...

CDI also offers a util class to easily implement this - [`AnnotationLiteral`](https://github.com/cdi-spec/cdi/blob/master/api/src/main/java/javax/enterprise/util/AnnotationLiteral.java).

I think we could easily leverage that and have a nice QoL improvement.
For instance user wanting to get hold of a named ME instance via `Instance` object in CDI could then do `Instance<Object>.select(ManagedExecutor.class, NamedInstance.NamedInstanceLiteral.of("myExec"))` as opposed creating another anonymous class to be able to reference the qualifier from code.

Will send PR shortly.

aguibert:
+1, I had just made a helper method for this in OpenLiberty but having a spec-provided literal creator method will be nice

aguibert:
to be consistent with other CDI literals, I think the name should just be `NamedInstance.Literal.of("myExec")`.  For example, all the qualifier annotations in `javax.enterprise.inject` follow the pattern of `<CLASS>.Literal.INSTANCE`

manovotn:
Not all of them, it varies based on whether the spec owns the annotation or not.
And ultimately it is a matter of choice, I am fine swapping it to just `Literal`.

PR #56, opened 2018-12-05
Issue #55 introduce annotation literals for ManagedExecutorConfig, Th…

manovotn:
…readContextConfig and NamedInstance.

Leverages CDI's `AnnotationLiteral` to implement helper classes for inline annotation creation.
Includes default instances for config annotations.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
@@ -27,6 +27,8 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
+import javax.enterprise.util.AnnotationLiteral;
+
 /**
  * <p>Annotates a CDI injection point for a {@link ManagedExecutor} such that the container
  * creates a new instance, which is identified within an application by its unique name.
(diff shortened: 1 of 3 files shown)


genie-microprofile:

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextConfig.java
```diff
@@ -165,4 +167,40 @@
...
      */
     String[] unchanged() default {};
+    
+        /**
+        * Util class used for inline creation of {@link ThreadContextConfig} annotation instances.
+        */
+        public final class ThreadContextConfigLiteral extends AnnotationLiteral<ThreadContextConfig> implements ThreadContextConfig {
+
+            public static final ThreadContextConfigLiteral DEFAULT_INSTANCE = 
+                of(new String[]{ThreadContext.TRANSACTION}, new String[]{ThreadContext.ALL_REMAINING}, new String[]{ThreadContext.ALL_REMAINING});
```
njr-11: There is a typo here specifying ALL_REMAINING for both the propagated value as well as unchanged.
Should be:
`of(new String[]{ThreadContext.TRANSACTION}, new String[]{}, new String[]{ThreadContext.ALL_REMAINING});`

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
```diff
@@ -164,4 +166,47 @@
      * <code>maxQueued</code> value is 0 or less than -1.
      */
     int maxQueued() default -1;
+
+        /**
+        * Util class used for inline creation of {@link ManagedExecutorConfig} annotation instances.
+        */
+        public final class ManagedExecutorConfigLiteral extends AnnotationLiteral<ManagedExecutorConfig> implements ManagedExecutorConfig {
```
aguibert: I think the class name can be reduced to just `Literal` on all of these, since repeating the class name is redundant: `ManagedExecutorConfig.ManagedExecutorConfigLiteral`.  Using just `Literal` would also be consistent with the CDI spec literal naming convention

aguibert: same comment also applies to the other literal classes introduced in this PR

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextConfig.java
manovotn: Good catch, thanks!

api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutorConfig.java
manovotn: Within CDI spec there are both variations, mostly depending on whether CDI owns that annotation or not. Ultimately it is a matter of choice. I am fine with just `Literal`.

manovotn:
@njr-11, @aguibert updated the PR accordingly

PR #58, opened 2018-12-07
Correct apache-rat-plugin now to blow up due to its own file.

manovotn:
When running `mvn clean install` the `rats` plugin blows up because it finds self-created text file with summary of license checks under spec/target and deems it incorrect.
I have adapted the exclusion list to ignore contents of target folders for when checking licenses.

I have also updated the plugin to its latest version while at it.


genie-microprofile:

manovotn:
@njr-11 Can we have Travis builds linked to PRs? It would eliminate trivial errors like this.

PR #59, opened 2018-12-07
Change ThreadContextConfig value property to propagated to align it w…

manovotn:
…ith what we have in builder.

@njr-11 I noted that the annotation uses `value` property, while on builder the same functionality is named `propagated`. I suppose that's incidental; this PR fixes it.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -149,7 +149,7 @@ capture and propagate Security context,
 
 [source,java]
 ----
-  @Inject @ThreadContextConfig(ThreadContext.SECURITY)
+  @Inject @ThreadContextConfig(propagated=ThreadContext.SECURITY)
   ThreadContext threadContext;
 ----
 
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/NamedInstance.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/NamedInstance.java
@@ -39,7 +39,7 @@ import javax.inject.Qualifier;
  * <pre><code> &commat;Inject &commat;NamedInstance("myExecutor") &commat;ManagedExecutorConfig(maxAsync=10)
  * ManagedExecutor myExecutor;
  *
- * &commat;Inject &commat;NamedInstance("myContext") &commat;ThreadContextConfig({ ThreadContext.SECURITY, ThreadContext.CDI })
+ * &commat;Inject &commat;NamedInstance("myContext") &commat;ThreadContextConfig(propagated = { ThreadContext.SECURITY, ThreadContext.CDI })
  * ThreadContext myThreadContext;
  * </code></pre>
  *
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextConfig.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContextConfig.java
@@ -39,7 +39,7 @@ import javax.enterprise.util.AnnotationLiteral;
  * <p>For example, the following injection points share a single
  * {@link ThreadContext} instance,</p>
  *
- * <pre><code> &commat;Inject &commat;NamedInstance("tc1") &commat;ThreadContextConfig({ ThreadContext.CDI, ThreadContext.APPLICATION })
+ * <pre><code> &commat;Inject &commat;NamedInstance("tc1") &commat;ThreadContextConfig(propagated = { ThreadContext.CDI, ThreadContext.APPLICATION })
  * ThreadContext threadContext1;
  *
  * &commat;Inject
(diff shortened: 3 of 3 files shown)


genie-microprofile:

manovotn:
Really?
Shorthand is one thing, but when I wanted to create a literal with custom values, I was puzzled for a while by what does `value` actually mean. You oftentimes want to specify more attributes and while they are all pretty self-explanatory, `value` sure isn't. Not unless you already know all three options that the annotation uses.

I can see the value of `value` (pardon the pun) when you have no other attribute in place. But with multiple, this is just confusing.

aguibert:
Perhaps as a compromise we could borrow an idea from [`@javax.servlet.annotation.WebServlet`](https://openliberty.io/docs/ref/javaee/8/#class=javax/servlet/annotation/WebServlet.html&package=javax/servlet/annotation/package-frame.html) and have both `value` and `propogated` available to users? Then we could simply doc the meaning of `value` as the same as `@see @ThreadContextConfig.propagated()`. 

For `@WebServlet` most of the time (>80%) the user will just want to specify the URL pattern like this:
```java
@WebServlet("MyServlet")
public class MyServlet extends HttpServlet {
```

But in some situations the user may want to specify more info, in which case the named attribute is available:
```java
@WebServlet(urlPatterns = {"MyServlet"}, description = "My Servlet")
public class MyServlet extends HttpServlet {
```

The `@WebServlet` doesn't javadoc the behavior, but if both `value` and `propagated` were specified it ought to be an error.

manovotn:
Perhaps that would be an option. Although we don't even have that in `ManagedExecutorConfig`, so it felt weird to put it here.
Also, `propagated` is set by default to `ALL_REMAINING` - you're not likely to change that most commonly. You can just alter the option (cleared) and leave this as is and it will still work which looks like yet another reason not to have it as `value`.

njr-11:
Having both `propagated`/`value` is not an option due to having a default value.  With both, it is impossible to distinguish between,
`@Inject @ThreadContextConfig(propagated = ThreadContext.CDI)`
and
`@Inject @ThreadContextConfig(propagated = ThreadContext.CDI, value = ThreadContext.ALL_REMAINING)`

I don't see the argument for `cleared` being a more common use case.  It is far more likely for an application to be able to fine-tune performance by telling you exactly the set of context it requires (propagated/value) than it is for an application to fine-tune performance by knowing of everything it doesn't want and telling you that (cleared).

aguibert:
Good point about default values, I had forgotten about that, so I agree having both attributes is not an option.

I am siding more with @manovotn on this one. Since the default value of `propagated/value` is `ALL_REMAINING`, it seems rare that a user would need to override this value.

Regarding this statement:
> I don't see the argument for cleared being a more common use case. It is far more likely for an application to be able to fine-tune performance by telling you exactly the set of context it requires (propagated/value) than it is for an application to fine-tune performance by knowing of everything it doesn't want and telling you that (cleared).

In this scenario, suppose that the user wants to propagate some custom `Foo` context that they care about.  You are saying they would specify their annotation as:
```java
@Inject
@ThreadContextConfig("Foo") // error... right?
ThreadContext fooCtx;
```

The values in the above case would be:
- propagated: Foo
- unchanged: <empty set>
- cleared: Transaction

However, wouldn't this be an error since other contexts such as `CDI` would be unspecified?  

---

If the user wants to fine-tune what contexts get propagated, I would think they must also override the cleared/unchanged value to set one of them to `ALL_REMAINING` like this:
```java
@Inject
@ThreadContextConfig(propagated = {"Foo"}, cleared = {ThreadContext.ALL_REMAINING})
ThreadContext fooCtx;
```

If my understanding is correct here, then I think it makes more sense to go with Matej's proposal of using `propagated` instead of `value`.

aguibert:
Ignore my previous comment, I had missed this critical bit of the `ThreadContext.Builder.propagated()` javadoc:

>         * Thread context types which are not otherwise included in this set or
>         * in the {@link #unchanged} set are cleared from the thread of execution
>         * for the duration of the action or task.</p>

Given this, I think the following scenario becomes more likely:
```java
@Inject
@ThreadContextConfig("Foo") // clears everything but 'Foo' context
ThreadContext fooCtx;```
```

IMO if we could establish that the above scenario was the dominant (>80% of the time) usage pattern, it would warrant using `value` on the annotation, but absent that, we should use `propagated`. 

Now that I understand this more, I don't feel strongly one way or another about `value` vs `propagated`.  I'm fine with whatever the two of you decide.

njr-11:
Between `value` vs `propagated`, we can go with the majority opinion here.

Using `value` streamlines the configuration for most usage because `cleared` and `unchanged` are more for special case, advanced usage.  When someone thinks of configuring thread context, what comes first to mind is the types of thread context they want, so it's very natural and intuitive to code something like,
`@Inject @ThreadContextConfig(ThreadContext.SECURITY)`
or
`@Inject @ThreadContextConfig({ThreadContext.APPLICATION, ThreadContext.CDI})`

Using `propagated` is not bad. It makes sense to the user and reads well, just extra typing and longer annotations to look at. I see the main benefit of `propagated` being that it makes the config a little bit clearer in the advanced cases, such as,
```
@Inject @ThreadContextConfig(propagated = {ThreadContext.APPLICATION, ThreadContext.CDI},
                             cleared = ThreadContext.SECURITY,
                             unchanged = ThreadContext.ALL_REMAINING)
```

The decision is really a trade-off over which scenarios we want to optimize for.

manovotn:
Updated

PR #61, opened 2018-12-12
create some fake thread context types for TCK tests

njr-11:
Create some fake thread context types for the TCK tests and use them to write a basic test confirming that third-party thread context providers which are made available via the ServiceLoader are found by the MicroProfile Concurrency implementation and used to capture/propagated/clear/restore thread context per the configuration that is supplied to the thread context builder.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/TckTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/TckTest.java
@@ -18,10 +18,22 @@
  */
 package org.eclipse.microprofile.concurrency.tck;
 
+import static org.eclipse.microprofile.concurrency.tck.contexts.priority.spi.ThreadPriorityContextProvider.THREAD_PRIORITY;
+
+import java.util.function.Supplier;
+
+import org.eclipse.microprofile.concurrency.tck.contexts.buffer.Buffer;
+import org.eclipse.microprofile.concurrency.tck.contexts.buffer.spi.BufferContextProvider;
+import org.eclipse.microprofile.concurrency.tck.contexts.label.Label;
+import org.eclipse.microprofile.concurrency.tck.contexts.label.spi.LabelContextProvider;
+import org.eclipse.microprofile.concurrency.tck.contexts.priority.spi.ThreadPriorityContextProvider;
+import org.eclipse.microprofile.concurrent.ThreadContext;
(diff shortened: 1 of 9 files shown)


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/TckTest.java
```diff
@@ -30,14 +42,94 @@
...
 
+    /**
+     * Verify that the MicroProfile Concurrency implementation finds third-party thread context providers
+     * that are made available to the ServiceLoader, allows their configuration via the ThreadContext builder,
+     * and correctly captures & propagates or clears these thread context types per the builder configuration.
+     * Subsequently verify that the MicroProfile Concurrency implementation properly restores thread context
+     * after the contextual action completes.
+     */
+    @Test
+    public void thirdPartyContextProvidersAreIncludedInThreadContext() {
```
aguibert: We should start splitting up TCK tests into different java classes, any categorization you come up with is fine by me, but we should have all the TCK tests for the spec in a single class.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/TckTest.java
```diff
@@ -30,14 +42,94 @@
 
     @Deployment
     public static WebArchive createDeployment() {
+        // build a JAR that provides fake 'ThreadPriority' context type
+        JavaArchive threadPriorityContextProvider = ShrinkWrap.create(JavaArchive.class, "threadPriorityContext.jar")
+                .addPackage(ThreadPriorityContextProvider.class.getPackage().getName())
```
aguibert: I would rather we hardcode the package name here instead of getting it from a class.  Looking at this code I don't know what package we are pulling in, and I'd need to hover over it in an IDE or expand imports to figure out what package we are pulling in.
Also, note that there is an `addPackages(boolean recursive, String... pkgNames)` which you may find useful.

njr-11: That makes sense, and I like the idea of using recursive to avoid needing to list all of the packages. I'll update it.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/TckTest.java
njr-11: I suppose it makes sense to do that from the start. I'll add a commit that splits it out.

PR #63, opened 2018-12-18
create ManagedExecutorTest for TCK and write tests for newIncompleteFuture, runAsync, and supplyAsync

njr-11:
Create the start of a TCK test class for ManagedExecutor.
Initially, include test for thread context propagation when using the following ManagedExecutor methods to create context-aware CompletableFutures:
newIncompleteFuture
runAsync
supplyAsync

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
@@ -262,16 +262,17 @@ public class ManagedExecutorTest extends Arquillian {
                 Assert.assertNull(v,
                         "Non-null value supplied to 'handle' method.");
 
-                // CompletableFuture JavaDoc is unclear on whether CompletionException or
-                // original exception is used here, so tolerate either.
-                if (x instanceof CompletionException && x.getCause() != null) {
-                    x = x.getCause();
-                }
+                Assert.assertEquals(x.getClass(), CompletionException.class,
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
```diff
@@ -0,0 +1,455 @@
...
+
+            Label.set("runAsync-test-label-E");
+
+            CompletableFuture<Character> stage5 = stage4.handle((v, x) -> {
+                Assert.assertNull(v,
+                        "Non-null value supplied to 'handle' method.");
+
+                // CompletableFuture JavaDoc is unclear on whether CompletionException or
+                // original exception is used here, so tolerate either.
+                if (x instanceof CompletionException && x.getCause() != null) {
```
nmittles: Java SE ComletableFuture behavior is to generate a CompletionException that wraps the original exception. We should change this test to be more restrictive to enforce that MP Concurrency has the same behavior.

njr-11: Good point - because ManagedExecutor returns a CompletableFuture (or extension thereof) here, we should validate behavior that remains consistent with Java SE's CompletableFuture class, even if CompletableFuture lacks clear documentation in this area.  This will guarantee the user the most consistent experience whether using Java SE CompletableFuture or CompletableFuture obtained from a ManagedExecutor.  I added a commit with the requested update.

Issue #64, opened 2018-12-18
ThreadContext.Builder and ManagedExecutor.Builder doc unclear on context set overlap


nmittles:
The JavaDoc for ThreadContext.Builder and ManagedExecutorBuilder are either incomplete or missing for defining behaviour related to overlap regarding the "cleared", "propagated", and "unchanged" thread context sets.

The JavaDoc should explicitly state that defining the same thread context type in more than one of "cleared", "propagated", or "unchanged" categories results in an error.

njr-11:
Thanks for catching this bug.  The equivalent CDI annotation classes ThreadContextConfig and ManagedExecutorConfig state this requirement, but it is missing from the builders.

PR #66, opened 2018-12-18
Issue #64: Update the ThreadContext.Builder JavaDoc for overlapping sets.

nmittles:
Removing the language around build failing for overlap between the cleared propagated and unchanged set from the cleared(), propagated(), and unchanged() methods. This error case is documented in the Builder.build() and is the method call that would actually generate the IllegalStateException.


genie-microprofile:

nmittles:
The PR fixes #64

Issue #60, opened 2018-12-12
Add TCK tests for MicroProfile Concurrency


njr-11:
Unless someone else knows of a better way to track this, I'm intending for this issue to be a place to document and collaborate on a list of test scenarios for the MicroProfile Concurrency TCK.

The following will be a checklist of scenarios. There are only a couple to start with, which seemed like some basic things we will need to start making progress. Sign up for a scenario to the TCK by mentioning the git pull number under which you are adding it. When the pull is merged, check the box next to it to indicate it is done,

- [x] #61 verify that the MP Concurrency impl provides a builder for ThreadContext
- [x] #68 verify ThreadContext.Builder API methods: cleared, propagated, unchanged
- [x] #103 verify ThreadContext builder is reusable
- [x] #69 verify ThreadContext API method: currentContextExecutor - runnables run with context
- [x] #62 verify ThreadContext API method: contextualCallable - callable runs with context
- [x] #62 verify ThreadContext API method: contextualConsumer - consumer runs with context
- [x] #62 verify ThreadContext API method: contextualFunction - function runs with context
- [x] #62 verify ThreadContext API method: contextualRunnable - runnable runs with context
- [x] #62 verify ThreadContext API method: contextualSupplier - supplier runs with context
- [x] #88, #108, #123 verify ThreadContext API method: withContextCapture - dependent actions run with context
- [x] #97 verify ThreadContext clears context types that are not configured under any category (per JavaDoc requirement)
- [x] #104, #112, #126 verify ThreadContext constants are defined and usable: TRANSACTION, SECURITY, CDI, ALL_REMAINING, and APPLICATION, to the extent that it is reasonable to do so when running the TCK.  Specifically, security context is left to the implementations to cover in their own testing due to the fact that it would introduce a dependency on servlets (HttpServletRequest.getUserPrincipal) or other vendor-specific technology that is not guaranteed to be available when running the TCK.
- [x] #61 verify that the MP Concurrency impl supports configuration and usage of third-party thread context provides supplied via the ServiceLoader
- [x] #114 The container creates a ThreadContext instance per unqualified ThreadContext injection point
- [x] #114 A ThreadContext instance that is injected by the container can be configured via `@ThreadContextConfig`
- [x] #114 If the `@NamedInstance` qualifier is present on a ThreadContext injection point that is annotated with `@ThreadContextConfig`, the container creates a ThreadContext instance that is qualified by the name value
[...]

PR #69, opened 2018-12-20
Issue 60: verify currentContextExecutor runs with context

nmittles:
Verify the ThreadContext API method currentContextExecutor generates an Executer that runs with context.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
@@ -583,7 +583,7 @@ public class ThreadContextTest extends Arquillian {
         }
     }
     
-    /**
+     /**
      * Verify the MicroProfile Concurrency implementation of propagate(), cleared(), and unchanged()
      * for ThreadContext.Builder.
      */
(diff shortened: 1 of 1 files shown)


nmittles:
This PR is part of the MicroProfile Concurrent TCK issue #60.

genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -580,4 +582,161 @@ public void thirdPartyContextProvidersAreIncludedInThreadContext() {
...
         }
     }
+    
+     /**
+     * Verify the MicroProfile Concurrency implementation of propagate(), cleared(), and unchanged()
+     * for ThreadContext.Builder.
+     */
+    @Test
+    public void contextControlsForThreadContextBuilder() throws InterruptedException, ExecutionException, TimeoutException {
+    	ThreadContext allContext = ThreadContext.builder()
```
njr-11: The name `allContext` is a bit misleading.  Intuitively, I would expect something named with all to propagate ThreadContext.ALL_REMAINING, clear nothing, and leave nothing unchanged.  A better name might be bufferContext, referring to what it does propagate.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -580,4 +582,161 @@ public void thirdPartyContextProvidersAreIncludedInThreadContext() {
...
+    	}
+
+        int originalPriority = Thread.currentThread().getPriority();
+        try {
+            // Set non-default values
+            int newPriority = originalPriority == 4 ? 3 : 4;
+            Buffer.get().append("contextControls-test-buffer-A");
+            Label.set("contextControls-test-label-A");
+
+            //Future<Integer> future1 = unmanagedThreads.submit(
```
njr-11: Commented out code is (accidentally?) left behind here

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -580,4 +582,161 @@ public void thirdPartyContextProvidersAreIncludedInThreadContext() {
...
+            	return Thread.currentThread().getPriority();
+            });
+            
+            Buffer.get().append("-B");
+            Label.set("contextControls-test-label-B");
+
+            Future<Integer> future = unmanagedThreads.submit(() -> {
+            		Buffer.get().append("unpropagated-buffer");
+            		Label.set("unpropagated-label");
+            		Thread.currentThread().setPriority(newPriority);
```
njr-11: You need a finally block to clean up each of these three contexts that you are setting on the unmanaged thread. Because it is an unmanaged thread, there is nothing to clean it up for you, so the context of this thread will remained corrupted when used by other tests at some future point.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -580,4 +582,161 @@ public void thirdPartyContextProvidersAreIncludedInThreadContext() {
...
+            });
+
+            Assert.assertEquals(future.get(MAX_WAIT_NS, TimeUnit.NANOSECONDS), Integer.valueOf(newPriority),
+                    "Callable returned incorrect value.");
+            
+            Assert.assertEquals(Buffer.get().toString(), "contextControls-test-buffer-A-B-C",
+            		"Context type was not propagated to contextual action.");
+
+            Assert.assertEquals(Label.get(), "contextControls-test-label-B",
+            		"Context type was not propagated to contextual action.");
```
njr-11: I think you forgot to update this failure message after copy/paste from elsewhere.  I'm not sure what to put for the message here though, because as far as I can tell it doesn't really test anything. This label was set onto the current thread earlier, but the current thread didn't do anything of consequence after that point which would cause it to change (all of the propagation work is done on a separate, unmanaged thread).

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -580,4 +582,161 @@ public void thirdPartyContextProvidersAreIncludedInThreadContext() {
...
+
+            Assert.assertEquals(Thread.currentThread().getPriority(), originalPriority,
+            		"Context type was not left unchanged on contextual action.");
```
njr-11: Similar to above comment.  It seems these sort of checks ought to instead be performed on the unmanaged thread from which you invoked the contextualCallable.call operation.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -580,4 +582,161 @@ public void thirdPartyContextProvidersAreIncludedInThreadContext() {
...
+    /**
+     * Verify that the MicroProfile Concurrency ThreadContext implementation's currentContextExecutor
+     * method can be used to create an Executor instance with the context that is captured from the
+     * current thread per the configuration of the ThreadContext builder, and that the context is
+     * applied to the thread where the Executor's execute method runs. This test case aligns with use 
+     * case of supplying a contextual Executor to a thread that is otherwise not context-aware.
+     */
+    @Test
+    public void currentContextExecutorRunsWithContext() throws InterruptedException, ExecutionException, TimeoutException {
+        ThreadContext priorityAndBufferContext = ThreadContext.builder()
```
njr-11: The name, priorityAndBufferContext, is inconsistent with the configuration that says it only propagates Buffer context.

Issue #74, opened 2019-01-02
ThreadContext#currentContextExecutorRunsWithContext TCK test question


manovotn:
Not sure if this is intentional, but [`ThreadContextTest`](https://github.com/eclipse/microprofile-concurrency/blob/master/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java#L680) at one point assumes that configuring TC builder with `propagated` only and then calling `currentContextExecutor()` will result in executor that clears all contexts that weren't specifically mentioned.

[This assertion](https://github.com/eclipse/microprofile-concurrency/blob/master/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java#L706-L707) even states (as fail msg) that "Context type that is configured to be cleared was not cleared." 

This IMO doesn't need to be the case because of our default values for undefined properties of TC. In case of this test, you will end up with:
_propagated - [Buffer]
cleared - [Transaction]
unchanged - []_

Is the test missing builder configuration for `cleared`, or am I missing something instead? :)
CCing @nmittles as he is the author.

njr-11:
I think I previously reviewed the pull that added this test, and I believe the test is valid as the [ThreadContext.Builder JavaDoc](https://github.com/eclipse/microprofile-concurrency/blob/master/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java) is currently written.  On lines 152-154 documenting the propagated() method, it states,
`Thread context types which are not otherwise included in this set or in the {@link #unchanged} set are cleared from the thread of execution for the duration of the action or task.`  The ThreadContextConfig.propagated method has a corresponding identical statement, and both ManagedExecutor.Builder.propagated and ManagedExecutorConfig.propagated have corresponding statements as well.

These statements (which are remnants of the design before we had the ALL_REMAINING constant) address the question of what to do with context types that are not specified in any list (cleared/propagated/unchanged).  For the sake of consistency across implementations, I believe it's a good idea for the spec to standardize on a behavior here, whether that be what is currently stated or otherwise if we think of something better.

The main alternative that I can think of is declaring it a configuration error (IllegalStateException upon build() and DefinitionException upon injection) when the user does not configure what to do with an available context type.  The main advantage here would be forcing the application to be explicit about what happens to every context type, such that the reader of the code will know exactly what happens to those other context types without consulting the JavaDoc.

[...]

PR #75, opened 2019-01-02
ThreadContextTest should declare Label as cleared context.

manovotn:


genie-microprofile:

njr-11:
Referencing my comment in #74, according to the current spec/JavaDoc, the change under this pull is valid, and so is the code prior to this pull.  Even if we don't end up changing the JavaDoc in any way, I'd still be happy with this pull given that it makes the test more clear, but I'll leave this open for a little bit longer just in case anyone else wants to comment.

PR #76, opened 2019-01-02
test maxAsync2 must not assume that ManagedExecutors will provide 2 threads to concurrently run async actions

njr-11:
The maxAsync2 TCK test is written in such a way that it assumes the ManagedExecutor can/will provide at least 2 threads, but there is no spec requirement for this.   If the ManagedExecutor only provides a single thread for async actions then the test ought to be considered as passing because the maxAsync=2 constraint will never be violated in that case.


manovotn:
I would not merge this; such change makes it a "volkswagen test" (always passing) which is pretty useless.
I am fine with the test as is, it actually pointed out mistake I made while initializing executor.
If, however, you want to change those tests, one way to address this could be to also define queue and fill up the executor + the queue. In such case any implementation 'should' stress the limits of executor and spawn `maxAsync` threads, right?

njr-11:
Discussed on the MP Concurrency call today that we will leave the test as is.  Expecting to be able run at least 2 async operations concurrently is a reasonable expectation to have of a ManagedExecutor.  If someone ever proposes an implementation that does otherwise (limits async tasks/actions to 1 or 0), we can update the spec/tests at that time to allow for that.

PR #77, opened 2019-01-03
Make sure we clear what we want cleared

FroMage:
This makes sure that we mark the contexts to clear, otherwise we're testing contexts for being cleared without reason.


FroMage:
Closing due to explanation in #74 about the surprising behaviour of `cleared`.

njr-11:
@FroMage the explanation in #74 doesn't make your pull invalid.  It would still be worth delivering for clarity of the tests.

FroMage:
Well it would improve clarity but would not test the implicit behaviour.

njr-11:
Yes, we would lose the coverage of the implicit clearing behavior here, but I added separate test scenarios under #60 to cover that behavior, so it would be okay.

PR #81, opened 2019-01-03
Do not specify that tasks should be cancelled on executor shutdown ca…

### LINKED ISSUE #78: State of Futures after shutting down executors
I was looking at current TCK tests and the docs we have around it.

[TCK seems to assume](https://github.com/eclipse/microprofile-concurrency/blob/master/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java#L963-L969) that all the references to `CompletableFuture` are in cancelled state after you shut down the executor. This however isn't what the spec says and IMO isn't even desirable behaviour because it goes against what other executors normally do. Standard behaviour is not to touch those and you can query your futures with timeouts or just check if `!f.isDone() && !f.isCancelled() && exec.isTerminated()` or maybe compare the runnables you passed there to those you got back from `shutdownNow()`.

As for ours docs, the only bit I found is a javadoc on MEBuilder (and similar piece on TCBuilder) which states:

> All created instances of ManagedExecutor are destroyed when the application is stopped. The container automatically shuts down these managed executors and **cancels their remaining actions/tasks**.

This bit speaks about what happens _when application is stopped_ at which point it doesn't really matter what state are the futures in as long as the executor shutdown is graceful which it is.

manovotn:
…ll. Alter relevant TCKs. Fixes #78.

@njr-11 this removes the bits from docs that state anything about cancellation (which only regards after application exits hence has no value) and also changes the test.
Although I suspect this way it will break your implementation which actually has the cancelled state, right? What change would be safe and still make the test sensible? Asserting just `!isDone()`?

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
@@ -109,7 +109,7 @@ public interface ManagedExecutor extends ExecutorService {
          *
          * <p>All created instances of {@link ManagedExecutor} are destroyed
          * when the application is stopped. The container automatically shuts down these
-         * managed executors and cancels their remaining actions/tasks.</p>
+         * managed executors.</p>
          *
          * @return new instance of {@link ManagedExecutor}.
          * @throws IllegalStateException for any of the following error conditions
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
@@ -82,10 +82,8 @@ public interface ThreadContext {
          *
          * <p>All created instances of {@link ThreadContext} are destroyed
          * when the application is stopped. The container automatically shuts down these
-         * {@link ThreadContext} instances, cancels their remaining
-         * <code>CompletableFuture</code>s and <code>CompletionStage</code>s, and
-         * and raises <code>IllegalStateException</code> to reject subsequent attempts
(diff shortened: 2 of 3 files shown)


njr-11:
I just saw this after adding a pull of my own (#82).  That is a good catch in ManagedExecutor/ThreadContext where you found some language where it was referring to cancelling actions/tasks.  I'm happy with that part.  Regarding the other changes you have to the test case, I think they went too far in testing the opposite behavior, which is also not guaranteed by the spec.  I'd like to use your changes to the JavaDoc and my updates to the test, which I think adequately allow for both possible behaviors - testing that the CompletableFutures do not report successful completion.  It should be a passing result for them to either be incomplete or to be complete due to cancellation/interruption, but not complete with successful result.

manovotn:
That's what I thought..
Your tests are pretty much no-op for any implementation that behaves differently to yours (e.g. that doesn't mark futures with runnables that weren't touched as done) which isn't good approach either.

Hence I'd try for a compromise. Maybe we could make a hybrid and assert both ways, even though then I am not sure it's viable TCK test, it sounds more a like a test tailored to pass both impls :-)
I will update this PR with separate commit and we can go from there.

One thing I found weird is that you consider it valid for `future1` to end with either `CancellationException` or `TimeoutException`. Timeout shouldn't really be the case as a task which was killed mid-way through execution should have some exceptional state but TO itself comes from just inability to grab the result in timely fashion. Cancellation is your impl choice so I added it there but as a user I would except different "state" on a task that was running a killed and a task that never started execution.

PR #82, opened 2019-01-03
TCK test cannot assume that shutdownNow will cause CompletableFutures to be canceled

njr-11:
The spec does not guarantee that CompletableFutures will be canceled upon shutdownNow just because the underlying executor cancels the tasks that would run those CompletableFutures, and therefore the TCK cannot require this behavior.  Another valid behavior would be for the CompletableFutures to remain in incomplete state.  This pull updates the test case to allow for either behavior, enforcing that CompletableFutures do not complete successfully in this case.


njr-11:
Closing this out with merging because the changes were included under #81

PR #85, opened 2019-01-04
write spec doc for integration with MP Config

### LINKED ISSUE #31: External configuration of what default context propagation means

njr-11:
pull fixes #31 

Write spec documentation for how MicroProfile Concurrency leverages MicroProfile Config to override configuration of instances produced by the container.

### DIFF
--- a/spec/src/main/asciidoc/mpconfig.asciidoc
+++ b/spec/src/main/asciidoc/mpconfig.asciidoc
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2018 Contributors to the Eclipse Foundation
+// Copyright (c) 2019 Contributors to the Eclipse Foundation
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -27,11 +27,9 @@ Applications can use MicroProfile Config in the standard way to enable configura
 ----
 @Produces @ApplicationScoped @NamedInstance("executor1")
 ManagedExecutor createExecutor(
-  @ConfigProperty(name="exec1.maxAsync", defaultValue="5") Integer a,
-  @ConfigProperty(name="exec1.maxQueued", defaultValue="20") Integer q,
-  @ConfigProperty(name="exec1.cleared", defaultValue=ThreadContext.TRANSACTION) String[] c,
-  @ConfigProperty(name="exec1.propagated", defaultValue=ThreadContext.ALL_REMAINING) String[] p) {
-    return ManagedExecutor.builder().maxAsync(a).maxQueued(q).propagated(p).cleared(c).build();
+    @ConfigProperty(name="exec1.maxAsync", defaultValue="5") Integer a,
+    @ConfigProperty(name="exec1.maxQueued", defaultValue="20") Integer q) {
+    return ManagedExecutor.builder().maxAsync(a).maxQueued(q).build();
 }
 ----
 
@@ -80,7 +78,7 @@ org.eclipse.microprofile.example.MyBean.exec4.maxAsync=7
 
 Note that `org.eclipse.microprofile.example.MyBean.exec4.maxAsync` overrides the configuration of the instance that is produced for injection into `exec4`. This same instance is injected into `exec5` per the matching `@NamedInstance("executor4")` qualifier.  It is an error to specify `org.eclipse.microprofile.example.MyBean.exec5.maxAsync`, which will not apply to `exec5`.
 
-To override a configuration attribute of an instance that is produced for injection into a parameter, specify a MicroProfile Config property with the name equal to the fully qualified class name, method name, parameter name, and configuration attribute name, delimited by the `.` character.
+To override a configuration attribute of an instance that is produced for injection into a parameter, specify a MicroProfile Config property with its name equal to the fully qualified class name, method name, parameter number (starting at 1), and configuration attribute name, delimited by the `.` character.
 
 [source, java]
 ----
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/mpconfig.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Inject @NamedInstance("executor6")
+    ManagedExecutor exec6;
+}
+----
+
+[source, text]
+----
+org.eclipse.microprofile.example.MyBean.createExecutor.exec.maxAsync=10
```
aguibert: We can't make this property include the parameter name, because method parameter names are only encoded into .class files if some special compiler flag is set.  Otherwise the parameter name at runtime will be `argN` where `N` is the position of the parameter.

I think we have two options here:
1. Use parameter position instead of parameter name, for example: `org.eclipse.microprofile.example.MyBean.createExecutor.0.maxAsync=10`
2. Automatically look for a parameter of type `ManagedExecutor` that is annotated with `@ManagedExecutorConfig`, which would allow us to simply reference the fully-qualified method name, for example:
`org.eclipse.microprofile.example.MyBean.createExecutor.maxAsync=10`
The only limitation with this approach is if a producer wanted to inject two `ManagedExecutor` parameters which were both annotated with `@MEC`. I think this is unlikely enough that it's a reasonable limitation. If someone did have this scenario they could always just look up the MP Config variables themselves.

spec/src/main/asciidoc/mpconfig.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+
+Applications can use MicroProfile Config in the standard way to enable configuration attributes of the `ManagedExecutor` and `ThreadContext` builders to be overriden.  For example,
+
+[source, java]
+----
+@Produces @ApplicationScoped @NamedInstance("executor1")
+ManagedExecutor createExecutor(
+  @ConfigProperty(name="exec1.maxAsync", defaultValue="5") Integer a,
+  @ConfigProperty(name="exec1.maxQueued", defaultValue="20") Integer q,
+  @ConfigProperty(name="exec1.cleared", defaultValue=ThreadContext.TRANSACTION) String[] c,
```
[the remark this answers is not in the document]
njr-11: Good point, there is no reason have this example be so complicated.  I'll switch to just maxAsync & maxQueued.

spec/src/main/asciidoc/mpconfig.asciidoc
njr-11: Good catch, I forgot about that detail.  I think with parameter position starting at 1 will make more sense than 0 based on the expectation that the user will count parameters and think in those terms.  Using the method name would be shorter, but also a bit awkward because the method itself doesn't have a maxAsync attribute.

spec/src/main/asciidoc/mpconfig.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+
+[source, text]
+----
+org.eclipse.microprofile.example.MyBean.exec2.maxAsync=5
+org.eclipse.microprofile.example.MyBean.exec3.maxAsync=6
+org.eclipse.microprofile.example.MyBean.exec4.maxAsync=7
+----
+
+Note that `org.eclipse.microprofile.example.MyBean.exec4.maxAsync` overrides the configuration of the instance that is produced for injection into `exec4`. This same instance is injected into `exec5` per the matching `@NamedInstance("executor4")` qualifier.  It is an error to specify `org.eclipse.microprofile.example.MyBean.exec5.maxAsync`, which will not apply to `exec5`.
+
```
Emily-Jiang: Why is it not allowed to specify the config value? IIRC, without the annotation `@ManagedExecutorConfig()`, the maxAsync is unlimited. You can configure it via a property? When you say "it is an error", do you stop the application from starting?

njr-11: Without the ManagedExecutorConfig annotation, with only the NamedInstance Qualifier, you do not get a new instance injected.  You get the existing instance that the Qualifier matches.  If you want to use MP Config to configure that instance, this is saying that it needs to be done based on the injection point for which that instance is created, not on one of the injection point with a Qualifier that uses it.  Anticipating that users might make this mistake, it was discussed on the last MP Concurrency call that it would be helpful to detect this and consider it an error rather than ignoring it.  A decision had not been reached on how to surface the error, so I didn't elaborate on that part in the doc.  Once a decision is reached, expect this paragraph to be updated with the outcome.

spec/src/main/asciidoc/mpconfig.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+To override a configuration attribute of an instance that is produced for injection into a parameter, specify a MicroProfile Config property with its name equal to the fully qualified class name, method name, parameter number (starting at 1), and configuration attribute name, delimited by the `.` character.
+
+[source, java]
+----
+package org.eclipse.microprofile.example;
+...
+public class MyBean {
+    @Produces @ApplicationScoped @NamedInstance("executor6")
+    ManagedExecutor createExecutor(@ManagedExecutorConfig(maxAsync=6) exec) {
```
manovotn: `@ManagedExecutorConfig(maxAsync=6) `*ManagedExecutor* `exec)`

njr-11: Thanks for catching the missing type in that example.  I've added a commit to fix it.

spec/src/main/asciidoc/mpconfig.asciidoc
njr-11: I'll assume the above answer was adequate, otherwise, we can include any further comments that you have on this paragraph under the issue where we document what the error path behavior will be, which is what it pertains to.

PR #87, opened 2019-01-08
Add clarification about behavior of programmatic CDI lookups

aguibert:


spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -99,6 +99,29 @@ public class BarBean {
...
+}
+----
+
+To perform an equivalent programmatic CDI lookup, the generated `@NamedInstance` qualifier must be taken into account. Namely:
+
+[source, java]
+----
+ManagedExecutor exec = CDI.current().select(ManagedExecutor.class, NamedInstance.Literal.of("com.foo.FooBean.exec")).get();
+----
+
```
njr-11: The above section seems like implementation detail that would create an awkward usage pattern for the user.  Can we omit that from the spec?  And then keep the following section which clarifies the behavior for programmatic lookup of unqualified instance

aguibert: Good point, that will cut down the size of this section considerably, and avoid exposing this possible implementation detail.

manovotn:
I know what are the pros of dynamic resolution, my question was more oriented towards what good is it, to try and resolve a `ManagedExecutor` without a name? As we have it, you cannot really use dynamic resolution to create a new instance (because we don't have a producer, user might though), you will have to grab an existing one. And all existing CDI beans of type ME have names (== qualifier) attached to them. Hence the dynamic resolution should IMO follow that and only allow you to retrieve beans with that qualifier.

Don't get me wrong, I am not against what you propose here, I simply don't see the value in it. It creates one extra instance of ME that lingers in no man's land - it has unconfigurable settings, no name and yet can effectively be shared if you use it from multiple places.

PR #88, opened 2019-01-08
TCK tests for ThreadContext.withContextCapture

njr-11:
Write TCK test for withContextCapture methods on ThreadContext.  These methods create a new CompletionStage or CompletableFuture based on one that is not context-aware, applying thread context propagation to dependent stages of the new instances (not the original).  Completion of the original instance causes the instances created by withContextCapture to complete, but not the other way around.  Also, cover later clarifications to the spec which indicate that new threads are not allocated for *Async operations because they are backed by a ThreadContext which can only run inline, rather than by a Managed Executor.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
@@ -23,6 +23,7 @@ import static org.eclipse.microprofile.concurrency.tck.contexts.priority.spi.Thr
 import java.lang.reflect.Method;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -582,7 +583,231 @@ public void thirdPartyContextProvidersAreIncludedInThreadContext() {
...
+            Label.set(null);
+        }
+    }
+
+    /**
+     * Verify that the MicroProfile Concurrency ThreadContext implementation's withContextCapture
+     * method can be used to create a dependent CompletableFuture instance that completes when the
+     * original stage completes and runs dependent stage actions with context that is captured
+     * from the thread that creates the dependent stage.
+     */
```
nmittles: Should the test description talk about testing a dependent CompletionStage rather than CompletableFuture?

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -582,7 +583,231 @@ public void thirdPartyContextProvidersAreIncludedInThreadContext() {
...
+        Assert.assertTrue(stage2.complete("stage_2_done"),
+                "It should be possible to complete a CompletableFuture created via withContextCapture without completing the original stage.");
+
+        Assert.assertFalse(stage1.isDone(),
+                "Completion of the dependent stage must not imply completion of the original stage.");
+
+        Assert.assertTrue(stage1.complete("stage_1_done"),
+                "It should be possible to complete the original stage with a different result after dependent stage was forcibly completed.");
+
+        Assert.assertEquals(stage1.get(), Integer.valueOf("stage_1_done"),
```
nmittles: Did you intend to use Integer.valueOf() here?

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -582,7 +583,231 @@ public void thirdPartyContextProvidersAreIncludedInThreadContext() {
...
+                "Completion stage result does not match the result with which it was forcibly completed.");
+
+        Assert.assertEquals(stage2.get(), Integer.valueOf("stage_2_done"),
```
nmittles: Did you intend to use Integer.valueOf() here?

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
njr-11: That's a copy/paste error and it seems this part of the test should never be able to pass.  I realized why - the test case is also missing the `@Test` annotation, so it wasn't running.  I'll add that as well when I correct this.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
njr-11: Fixing, the same as above

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
njr-11: Yes, that needs to be updated

PR #89, opened 2019-01-09
Add tests for CDI injection of ManagedExecutor

aguibert:


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
```diff
@@ -0,0 +1,170 @@
...
+    @NamedInstance("appProduced")
+    public ManagedExecutor createExec() {
+        return ManagedExecutor.builder().build();
+    }
+
+    @Produces
+    @ApplicationScoped
+    @NamedInstance("appProduced_injected")
+    public ManagedExecutor createExecInjected(@ManagedExecutorConfig(maxAsync = 2) ManagedExecutor exec) {
+        System.out.println("@AGG in producer with exec: " + exec);
```
njr-11: println and dumpStack accidentally left in code?

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
```diff
@@ -0,0 +1,170 @@
...
+        Thread.dumpStack();
+        return exec;
+    }
+
+    /**
+     * Verify that injected ME instances injected by the container can be shutdown
+     */
+    public void shutdownContainerInstance() throws Exception {
```
njr-11: Does this test run? I don't see anything that invokes it.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
```diff
@@ -0,0 +1,170 @@
...
+        throwAway1.shutdown();
+        assertTrue(throwAway1.awaitTermination(5, TimeUnit.SECONDS));
```
njr-11: Recommend using the same default maximum allowable timeout as other test cases in the MP Concurrency bucket.  Also applies a few lines later

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * Verify that injected ME instances are useable in a very basic way
+     */
+    public void testBasicExecutorUsable() throws Exception {
+        assertEquals(exec.supplyAsync(() -> "hello").get(5, TimeUnit.SECONDS), "hello");
+        assertEquals(maxAsync2.supplyAsync(() -> "hello").get(5, TimeUnit.SECONDS), "hello");
+        assertEquals(appProduced.supplyAsync(() -> "hello").get(5, TimeUnit.SECONDS), "hello");
+        assertEquals(appProduced_injected.supplyAsync(() -> "hello").get(5, TimeUnit.SECONDS), "hello");
```
njr-11: Again, let's be consistent in maximum wait/timeout usage with the other test cases

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * Verify that the @ManagedExecutorConfig annotation is applied to injection points
+     */
+    public void testMaxAsync2() throws Exception {
```
njr-11: This is really 2 different tests.  They should not be combined.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
```diff
@@ -0,0 +1,170 @@
...
+        checkMaxAsync2(maxAsync2);
+        checkMaxAsync2(appProduced_injected);
+    }
+    
+    private void checkMaxAsync2(ManagedExecutor exec) throws Exception {
```
njr-11: Verifying maxAsync is less reliable than verifying maxQueued because it involves waiting for a fixed amount of time and observing whether additional tasks might have started.  This is both wasteful of test execution time and not guaranteed to actually catch errors.  I understand why we have one test that does this (which I wrote in a different test class and which this partly seems to copy from) because we want at least some amount of verification.  But I don't see why we would want to repeat this test multiple times as a way of verifying that `@ManagedExecutorConfig` in general was processed.  This should not be a utility method for repeated usage.  It should only be done once.  To verify the general case that `@ManagedExecutorConfig` was applied, it would be better to use maxQueued, propagated, or cleared.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/ManagedExecutorSharingBean.java
```diff
@@ -0,0 +1,114 @@
...
+    public void testUnnamedConfigDifferent() {
+        assertNotSame(defaultAnno1, defaultAnno2,
+                "Default injection points with matching @ManagedExecutorConfig should get different instances");
+    }
+
+    /**
+     * Verify injection points with matching @NamedInstance qualifiers share an ME instance
+     */
+    public void testNamedExecsSame() {
+        assertEquals(namedExecA1, namedExecA2,
```
njr-11: This isn't really testing that the injection points share an instance, just that the instances they return are .equals to each other.  This should be switched to assertSame

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/ManagedExecutorSharingTest.java
```diff
@@ -0,0 +1,68 @@
...
+    @Deployment
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class, ManagedExecutorSharingTest.class.getSimpleName() + ".war")
+                .addPackage("org.eclipse.microprofile.concurrency.tck.cdi")
+                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    @Test
+    public void testDefaultMEsDifferent() {
+        testBean.testDefaultMEsDifferent();
```
manovotn: Again, I'd start with null checks.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/ManagedExecutorSharingBean.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Arrays;
+
+import javax.enterprise.context.RequestScoped;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.concurrent.ManagedExecutor;
+import org.eclipse.microprofile.concurrent.ManagedExecutorConfig;
+import org.eclipse.microprofile.concurrent.NamedInstance;
+
+@RequestScoped
```
manovotn: Same as above, application scoped would be better choice.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/BasicCDITest.java
```diff
@@ -0,0 +1,59 @@
...
+    @Deployment
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class, BasicCDITest.class.getSimpleName() + ".war")
+                .addPackage("org.eclipse.microprofile.concurrency.tck.cdi")
+                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    @Test
+    public void testVerifyInjection() {
+        bean.testVerifyInjection();
```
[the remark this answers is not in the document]
aguibert: if something goes wrong and the test harness doesn't inject the bean, the test would still fail with a NPE. I can add a check for this test though

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
```diff
@@ -0,0 +1,170 @@
...
+import javax.enterprise.context.RequestScoped;
+import javax.enterprise.inject.Produces;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.concurrent.ManagedExecutor;
+import org.eclipse.microprofile.concurrent.ManagedExecutorConfig;
+import org.eclipse.microprofile.concurrent.NamedInstance;
+import org.testng.Assert;
+
+@RequestScoped
```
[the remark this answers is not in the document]
aguibert: I'm running in the Liberty container. Originally I was going to do a multi-request test case to verify that the same ManagedExecutor was returned on multiple requests, but that ended up being way too complex in Arquillian. I'll change to this `@AppScoped`

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
aguibert: oops!

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
aguibert: nope, good catch

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
aguibert: good point, I'll use maxQueued instead since that is more efficient

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/ManagedExecutorSharingBean.java
aguibert: both good points, do we think it is sufficient to just verify that the ManagedExecutor toString()'s are equal?
It looks like the proxies do return the toString of the underlying objects -- for example:
`com.ibm.ws.concurrent.mp.ManagedExecutorImpl@66ab0056 `

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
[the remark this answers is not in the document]
manovotn: The test you describe would be easier to achieve in pure SE where you can manually activate/deactivate req. context. With Arq., a lot of adapters tend to activate contexts before test starts, which makes this impossible.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/ManagedExecutorSharingBean.java
manovotn: Yes I think `toString()` will work so long as implementation is requires to override it (which is one of the reasons why I originally wanted to have that) and we already have a PR for that.
As I stated in https://github.com/eclipse/microprofile-concurrency/pull/92#discussion_r249139741, CDI forwards `toString()` invocations to the actual underlying instance so long as the implementation overrides it. What I forgot to add is that since that time I verified it manually as well :)

PR #92, opened 2019-01-11
requirements of toString for ManagedExecutor and ThreadContext

njr-11:
Include requirements for toString to show configuration detail and injection point name (if container creates it for CDI injection).

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018,2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -313,4 +313,14 @@ public interface ManagedExecutor extends ExecutorService {
      * @return the new completion stage.
      */
     <U> CompletableFuture<U> supplyAsync(Supplier<U> supplier);
+
+    /**
+     * <p>Returns a textual representation of this executor. For managed executors
+     * that are produced by the container for CDI injection, the textual representation must include
+     * the fully qualified name of the injection point from which the container produces the
+     * managed executor.</p>
+     *
+     * @return a textual representation of the executor.
+     */
+    String toString();
 }
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018,2019 Contributors to the Eclipse Foundation
  *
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
```diff
@@ -298,4 +298,34 @@ public static Builder builder() {
      * @return the new completion stage.
      */
     <U> CompletableFuture<U> supplyAsync(Supplier<U> supplier);
+
+    /**
+     * <p>Returns a textual representation of this executor, including detail about its
+     * configuration attributes.</p>
+     *
+     * <p>For managed executors that are created by the container for CDI injection
```
manovotn: Do we need to be this precise? It's quite a wall of text for "just `toString()`".

Why not just say something like:
`For managed executors that are created by the container for CDI injection, the method also includes their unique name.`

manovotn: I recall you were against leveraging `toString()` for TCK comparisons. Though the next best bet I have is to depend (in tests) on Weld API to get the actual instance.

Either way, the level of detail depends on how you wish to make comparison with it. Whether it's the precise form, or whether you would simply use is to compare two executors as `ex1.toString().equals(ex2.toString())` for no matter the actual form implementation chooses, the comparison should return true only if they have the same settings and name. In CDI that would mean equivalence of those two executors.

manovotn: And to answer your question - yes, it's dependent on how Andy decides to re-design those tests.

njr-11: I'm not against the usage toString by the TCK, so long as it is sufficiently well-documented to make it valid to do so, which is what we are trying to accomplish here.  One question would be whether `equals()` comparison on toString results is safe.  Would two proxied instances ever return something like
`ManagedExecutorProxy@11112222_org.eclipse.example.MyBean.executor1`
and
`ManagedExecutorProxy@33334444_org.eclipse.example.MyBean.executor1`
?
That is what I was worried about when I put the detail about how to precisely form the unique injection point name, because then tests could assert that exec1.contains(computedUniqueName) and exec2.contains(computedUniqueName).  But if the proxies won't interfere with the toString, then the approach you suggested is fine and we can remove all of the precise language around how the injection point name is formed.

manovotn: >  But if the proxies won't interfere with the toString

Yea, that's exactly what I meant. Should be easy enough to check, I am not 100% sure from the top of my head, but I think the invocation of `toString()` on a client proxy should be forwarded to the actual bean (if it overrides it OFC). I mean if it didn't then it'd be pretty useless to define it with proxies in play.

This bit of specification seem to support it - http://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#client_proxy_invocation
Although there doesn't seem to be any TCK test for this (did just a quick check).

njr-11: Per review comments above, I have replaced the commit under this pull with one that makes a simpler requirement only to include the name in the toString output.

PR #96, opened 2019-01-17
TCK test where MP Config overrides injection point with parameter position > 1

njr-11:
This pull adds at least one TCK test that covers MicroProfile Config override of configuration for a parameter injection point other than the first parameter.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
@@ -423,7 +423,7 @@ public class MPConfigTest extends Arquillian {
 
         // Expected config is propagated=Label; unchanged=ThreadPriority, cleared=Remaining
         Assert.assertNotNull(producedThreadContext,
-                "Unable to inject ThreadContext into method parameter 3. Cannot run test.");
+                "Unable to inject ThreadContext. Cannot run test.");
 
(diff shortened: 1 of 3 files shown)


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
```diff
@@ -406,6 +412,40 @@ public void overrideContextPropagationForThreadContextParameterWithConfig() {
...
+     * of a ThreadContext that is produced by the container when the application defines
+     * an unqualified parameter injection point that is annotated with ThreadContextConfig,
+     * where the ThreadContext parameter is the third parameter of the method.
+     */
+    @Test(dependsOnMethods = "beanInjected")
+    public void overrideContextPropagationForThreadContextParameter3WithConfig() {
+
+        // Expected config is propagated=Label; unchanged=ThreadPriority, cleared=Remaining
+        Assert.assertNotNull(producedThreadContext,
+                "Unable to inject ThreadContext into method parameter 3. Cannot run test.");
```
nmittles: Parameter's 1 & 2 might also cause problems, and this message could then be misleading.

njr-11: Added a commit which makes the failure message more generic to cover other possible causes.  Now it just says unable to inject.

PR #97, opened 2019-01-17
Add TCK testing of ThreadContext for implied clearing of contexts.

nmittles:
Write TCK tests covering the ThreadContext.Builder.cleared() API will implicitly include all remaining contexts there were not otherwise specified. Including testing with MicroProfile Config to create a scenario where the cleared list should be updated through an indirect change of the propagated list.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
@@ -24,8 +24,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.Phaser;
 import java.util.concurrent.RejectedExecutionException;
@@ -415,15 +413,11 @@ public class MPConfigTest extends Arquillian {
     /**
      * Verify MicroProfile Config overrides the propagated attribute to enabled
      * an implied clear of all remaining.
-     * @throws TimeoutException 
-     * @throws ExecutionException 
-     * @throws InterruptedException 
+     * @throws Exception 
      */
     @Test
-    public void overrideContextPropagationForThreadContextWithImpliedCleared() throws InterruptedException, ExecutionException, TimeoutException {
-        ExecutorService unmanagedThreads = null;
-        
-        // Expected config is propagated=Buffer; cleared=Remaining; unchanged=Label
+    public void overrideContextPropagationForThreadContextWithImpliedCleared() throws Exception {
+        // Expected config is propagated=Buffer; cleared=Transaction(with implied Remaining); unchanged=Label
         Assert.assertNotNull(clearAllRemainingThreadContext,
                 "Unable to inject ThreadContext qualified by NamedInstance. Cannot run test.");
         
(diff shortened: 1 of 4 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
```diff
@@ -405,6 +411,81 @@ public void overrideContextPropagationForThreadContextParameterWithConfig() {
...
+     * an implied clear of all remaining.
+     * @throws TimeoutException 
+     * @throws ExecutionException 
+     * @throws InterruptedException 
+     */
+    @Test
+    public void overrideContextPropagationForThreadContextWithImpliedCleared() throws InterruptedException, ExecutionException, TimeoutException {
+        ExecutorService unmanagedThreads = null;
+        
+        // Expected config is propagated=Buffer; cleared=Remaining; unchanged=Label
```
njr-11: Based on its absence from MP Config and default value for cleared, I would expect the config to be cleared=Transaction here.  And then, given that none of the config specifies ALL_REMAINING, this implies that all remaining context is cleared.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
```diff
@@ -405,6 +411,81 @@ public void overrideContextPropagationForThreadContextParameterWithConfig() {
...
+            Buffer.set(new StringBuffer("clearUnspecifiedContexts-test-buffer-A"));
+            Label.set("clearUnspecifiedContexts-test-label-A");
+
+            Callable<Integer> callable = clearAllRemainingThreadContext.contextualCallable(() -> {
+                try {
+                    Assert.assertEquals(Buffer.get().toString(), "clearUnspecifiedContexts-test-buffer-A",
+                            "Context type was not propagated to contextual action.");
+
+                    Assert.assertEquals(Label.get(), "",
+                            "Context type was not left unchanged by contextual action.");
```
njr-11: Using an executor service to run on a thread with a cleared Label isn't a good combination for this test because it prevents you from fully testing that Label is unchanged per the configuration.  There should be no need to create an executor service here, because it also makes the test unnecessarily complicated as well as interfering here with verifying results.  Instead, once the contextualCallable method finishes, you can update the context on the main test thread to whatever you want before running it.  For example, to clearUnspecifiedContexts-test-buffer-C and clearUnspecifiedContexts-test-label-C.  Then run the contextualCallable inline, which will let you verify that the label was left unchanged on the thread.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
```diff
@@ -405,6 +411,81 @@ public void overrideContextPropagationForThreadContextParameterWithConfig() {
...
+
+                    Label.set("clearUnspecifiedContexts-test-label-B");
+
+                    return Thread.currentThread().getPriority();
+                }
+                finally {
+                    // Restore original values
+                    Buffer.set(null);
+                    Label.set(null);
+                    Thread.currentThread().setPriority(originalPriority);
```
njr-11: What is the point of having the contextual task restore values itself?  It is the job of the MP Concurrency implementation to do this, and the values here just interfere with ensuring that is done properly.  Also note that you are completely undoing here what you did on line 446 without testing anything related to it in between.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
```diff
@@ -405,6 +411,81 @@ public void overrideContextPropagationForThreadContextParameterWithConfig() {
...
+                }
+            });
+
+            unmanagedThreads = Executors.newFixedThreadPool(5);
```
njr-11: As stated earlier, I think the extra thread pool gets in the way of the test logic rather than helping it.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
```diff
@@ -405,6 +411,81 @@ public void overrideContextPropagationForThreadContextParameterWithConfig() {
...
+            Future<Integer> future = unmanagedThreads.submit(() -> {
+                try {
+                    Thread.currentThread().setPriority(newPriority);
```
njr-11: setting the priority here would be completely unnecessary if following the approach of running on the main test thread

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
```diff
@@ -405,6 +411,81 @@ public void overrideContextPropagationForThreadContextParameterWithConfig() {
...
+                    Label.set(null);
+                    Thread.currentThread().setPriority(originalPriority);
+                }
+            });
+
+            Assert.assertEquals(future.get(MAX_WAIT_NS, TimeUnit.NANOSECONDS), Integer.valueOf(Thread.NORM_PRIORITY),
+                    "Context type that remained unspecified was not cleared by default.");
+            
+            Assert.assertEquals(Label.get(), "clearUnspecifiedContexts-test-label-A",
+                    "Context type was not left unchanged by contextual action.");
```
njr-11: This doesn't test anything useful due to the separate thread being used.  However, if you remove that, and all the manual clearing logic, then you could test here that the label matches clearUnspecifiedContexts-test-label-B, meaning that the MP Concurrency implementation left it alone (the intent of `unchanged`)

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -116,6 +116,75 @@ public void builderForThreadContextIsProvided() {
...
+    
+    /**
+     * Verify that the MicroProfile Concurrency ThreadContext implementation clears context
+     * types that are not configured under propagated, unchanged, or cleared.
+     * @throws TimeoutException 
+     * @throws ExecutionException 
+     * @throws InterruptedException 
+     */
+    @Test
+    public void clearUnspecifiedContexts() throws InterruptedException, ExecutionException, TimeoutException {
```
njr-11: This seems identical to the prior test (excepting that builder is used instead of CDI/MPConfig) so the same comments apply.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
```diff
@@ -405,6 +409,58 @@ public void overrideContextPropagationForThreadContextParameterWithConfig() {
...
+            });
+
+            Buffer.set(new StringBuffer("clearUnspecifiedContexts-test-buffer-C"));
+            Label.set("clearUnspecifiedContexts-test-label-C");
+
+            Assert.assertEquals(callable.call(), Integer.valueOf(Thread.NORM_PRIORITY),
+                    "Context type that remained unspecified was not cleared by default.");
+            
+            Assert.assertEquals(Buffer.get().toString(), "clearUnspecifiedContexts-test-buffer-C",
+                    "Context type was not left unchanged by contextual action.");
```
njr-11: This context type isn't left unchanged. It is propagated and then restored afterward. So the test here is correct, but the messages is wrong.  Should be "Context type was not restored after contextual action."

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -116,6 +116,59 @@ public void builderForThreadContextIsProvided() {
...
+            });
+
+            Buffer.set(new StringBuffer("clearUnspecifiedContexts-test-buffer-C"));
+            Label.set("clearUnspecifiedContexts-test-label-C");
+
+            Assert.assertEquals(callable.call(), Integer.valueOf(Thread.NORM_PRIORITY),
+                    "Context type that remained unspecified was not cleared by default.");
+            
+            Assert.assertEquals(Buffer.get().toString(), "clearUnspecifiedContexts-test-buffer-C",
+                    "Context type was not left unchanged by contextual action.");
```
njr-11: same comment as above

PR #98, opened 2019-01-18
Fix some specification text errors

xstefank:
- code escape overrun in concurrencyprovider.asciidoc
- missing `@` for ThreadContextConfig in overview.asciidoc
- adjusted application MP config example to use only properties visible in the example in mpconfig.asciidoc


genie-microprofile:

PR #100, opened 2019-01-18
Add TCK testing of ManagedExecutor for implied clearing of contexts.

nmittles:
Write TCK tests covering the ManagedExecutor.Builder.cleared() API will implicitly include all remaining contexts there were not otherwise specified.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
@@ -158,6 +158,7 @@ public class ManagedExecutorTest extends Arquillian {
                     "Previous context (Buffer) was not restored after context was propagated for contextual action.");
         }
         finally {
+            executor.shutdownNow();
             // Restore original values
             Buffer.set(null);
             Thread.currentThread().setPriority(originalPriority);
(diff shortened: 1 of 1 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
```diff
@@ -121,6 +122,48 @@ public void builderForManagedExecutorIsProvided() {
...
+                    Assert.assertEquals(Buffer.get().toString(), "clearUnspecifiedContexts-test-buffer-A",
+                            "Context type was not propagated to contextual action.");
+
+                    Buffer.set(new StringBuffer("clearUnspecifiedContexts-test-buffer-B"));
+
+                    return Thread.currentThread().getPriority();
+            });
+
+            Assert.assertEquals(future.get(MAX_WAIT_NS, TimeUnit.NANOSECONDS), Integer.valueOf(Thread.NORM_PRIORITY),
+                    "Context type that remained unspecified was not cleared by default.");
```
njr-11: In order to truly test that context type is cleared, use executor.completedFuture(x).thenRun(task), such that it will need to clear the context that is already established on the main test thread rather than getting a pooled thread which starts out cleared where it would be a no-op.

PR #101, opened 2019-01-18
Add section on MP Config overrides for array value properties, clarifying how to configure empty values

### LINKED ISSUE #99: Document how to use MP Config to override context list with empty
Our documentation for overrides via MicroProfile Config neglects to document how to use MicroProfile Config to clear our array properties (propagated context, cleared context, unchanged context).  It turns out that MicroProfile Config is unclear on exactly how to do this.  I've opened https://github.com/eclipse/microprofile-config/issues/397 against MicroProfile Config for this.  The main problem is disambiguating between an empty array and an array of empty string.  A property such as `org.eclipse.examples.MyBean.threadContext.unchanged=` could be interpreted either way.  In order for our users to have a solution, we can accommodate this by stating that both shall be interpreted by MicroProfile Concurrency to have the same meaning of indicating no context types.  Thus, the user will be able to rely on specifying the blank value regardless of which way MP Config supplies that value to the MP Concurrency implementation.

njr-11:
pull fixes #99 

This pull provides an example of using MicroProfile Config to override array properties (the lists of context types) and documents a requirement that MicroProfile Concurrency providers will need to be aware of to cope with an ambiguity in MicroProfile Config around empty values in a consistent manner.

### DIFF
--- a/spec/src/main/asciidoc/mpconfig.asciidoc
+++ b/spec/src/main/asciidoc/mpconfig.asciidoc
@@ -119,7 +119,7 @@ package org.eclipse.microprofile.example;
 ...
 public class MyBean {
     @Inject
-    ThreadContext appAndCDIContextPropagator;
+    ThreadContext contextPropagator;
 }
 ----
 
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/mpconfig.asciidoc
```diff
@@ -101,3 +101,35 @@ org.eclipse.microprofile.example.MyBean.createExecutor.1.maxAsync=10
...
+If the user wishes to override the above to propagate exactly 2 context types (`Application` and `CDI`), clear nothing, and leave all `Remaining` types unchanged, then they could configure the following MicroProfile Config properties,
+
+[source, text]
+----
+org.eclipse.microprofile.example.MyBean.appAndCDIContextPropagator.propagated=Application,CDI
+org.eclipse.microprofile.example.MyBean.appAndCDIContextPropagator.cleared=
+org.eclipse.microprofile.example.MyBean.appAndCDIContextPropagator.unchanged=Remaining
+----
+
+In order to guarantee that empty string config values are interpreted properly, the MicroProfile Concurrency implementation must accept both an empty array or collection as well as a size 1 array or collection containing the empty string as meaning empty.  This is necessary due to a lack of clarity in the first several versions of the MicroProfile Config specification about how the empty string value is to be interpreted for arrays and collections of String.
```
manovotn: I would structure this will bullets to make it more readable. Probably something like this:

_In order to guarantee that empty string config values are interpreted properly, the MicroProfile Concurrency implementation must interpret the following scenarios as empty:_
* _empty array_
* _an array containing the empty string as its singular element_

_This is necessary due to ..._


As a side note, why did you mention collections specifically? It's just arrays, isn't it?
> `...must accept both an empty array or collection...`

spec/src/main/asciidoc/mpconfig.asciidoc
```diff
@@ -101,3 +101,35 @@ org.eclipse.microprofile.example.MyBean.createExecutor.1.maxAsync=10
...
+
+For example, we can start with the following injection point, which has the default configuration for `ThreadContext`. This means clearing the `Transaction` context, leaving no context type unchanged, and propagating all `Remaining` context types,
+
+[source, java]
+----
+package org.eclipse.microprofile.example;
+...
+public class MyBean {
+    @Inject
+    ThreadContext appAndCDIContextPropagator;
```
manovotn: I would choose more neutral field name(`configurableContextPropagator`?), this one can be confusing as from the code itself it looks like it has default settings whereas the actual behaviour depends on the settings described below.

spec/src/main/asciidoc/mpconfig.asciidoc
njr-11: Good idea, I'll update the formatting as suggested.
Also true regarding array vs collections. The annotation only uses array, so there is no need to mention collections.

spec/src/main/asciidoc/mpconfig.asciidoc
njr-11: Yes, I can update that name to make it more generic

PR #102, opened 2019-01-18
Add TCK testing of ManagedExecutor for reusing Builder.build()

nmittles:
Write TCK tests covering that ManagedExecutor.Builder.builder() can be reused with different context configurations.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
@@ -1242,17 +1242,9 @@ public class ManagedExecutorTest extends Arquillian {
         
         ManagedExecutor clearingExecutor = builder.build();
 
-        builder.propagated(Buffer.CONTEXT_NAME);
-        try {
-            builder.build();
-            Assert.fail("ManagedExecutor.Builder.build() should throw an IllegalStateException for set overlap between propagated and cleared");
-        }
-        catch (IllegalStateException ISE) {
-            // test passes
-        }
-
-        builder.cleared();
-        ManagedExecutor propagatingExecutor = builder.build();
+        ManagedExecutor propagatingExecutor = builder.propagated(Buffer.CONTEXT_NAME)
+                .cleared()
+                .build();
 
         try {
             // Set non-default value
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
```diff
@@ -1230,6 +1230,65 @@ public void shutdownPreventsAdditionalSubmits() throws ExecutionException, Inter
...
+            Future<Void> clearedFuture = clearingExecutor.completedFuture(1).thenRun(() -> {
+                Assert.assertEquals(Buffer.get().toString(), "",
+                        "Context type that is configured to be cleared was not cleared.");
+
+                Buffer.set(new StringBuffer("reuseBuilder-test-buffer-B"));
+            });
+
+            Future<Void> propagatedFuture = propagatingExecutor.completedFuture(1).thenRun(() -> {
+                Assert.assertEquals(Buffer.get().toString(), "reuseBuilder-test-buffer-A",
+                        "Context type was not propagated to contextual action.");
```
njr-11: This task should use thenRunAsync instead of running it inline where the context to propagate will already be on the test's main thread.

nmittles: Good point. Updating...

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ManagedExecutorTest.java
```diff
@@ -1230,6 +1230,65 @@ public void shutdownPreventsAdditionalSubmits() throws ExecutionException, Inter
...
+        ManagedExecutor.Builder builder = ManagedExecutor.builder()
+                .propagated()
+                .cleared(Buffer.CONTEXT_NAME);
+        
+        ManagedExecutor clearingExecutor = builder.build();
+
+        builder.propagated(Buffer.CONTEXT_NAME);
+        try {
+            builder.build();
+            Assert.fail("ManagedExecutor.Builder.build() should throw an IllegalStateException for set overlap between propagated and cleared");
```
manovotn: I think this is already part of another test dedicated to it.
It also looks weird to first change the state to something you know should blow up and then still use the same instance (which you deliberately shifted to "broken state") to build something valid.

PR #103, opened 2019-01-18
Add TCK testing of ThreadContext for reusing Builder.build()

nmittles:
Write TCK tests covering that ThreadContext.Builder.builder() can be reused with different context configurations.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
@@ -626,6 +626,7 @@ public class ThreadContextTest extends Arquillian {
             });
             
             Buffer.set(new StringBuffer("reuseBuilder-test-buffer-D"));
+            Thread.currentThread().setPriority(newPriority - 1);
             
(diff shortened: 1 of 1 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ThreadContextTest.java
```diff
@@ -581,6 +581,67 @@ public void contextualSupplierRunsWithContext() throws InterruptedException {
...
+                Assert.assertEquals(Buffer.get().toString(), "reuseBuilder-test-buffer-A",
+                        "Context type was not propagated to contextual action.");
+
+                Buffer.set(new StringBuffer("reuseBuilder-test-buffer-C"));
+                
+                return Thread.currentThread().getPriority();
+                
+            });
+            
+            Buffer.set(new StringBuffer("reuseBuilder-test-buffer-D"));
```
njr-11: I think we should also change the thread priority here to verify that the previously captured 'thread priority' context is actually propagated rather than just reporting what is already on the thread

nmittles: Will update shortly.

PR #105, opened 2019-01-22
Add TCK testing of custom ConcurrencyManager instances.

nmittles:
Test obtaining a ConcurrencyManager builder from the ConcurrencyProvider. Then validate building, registering, and releasing a custom ConcurrencyManager. If a ConcurrencyManager builder is not supported then verify an UnsupportedOperationException is thrown.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ConcurrencyManagerTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ConcurrencyManagerTest.java
@@ -62,32 +62,54 @@ public class ConcurrencyManagerTest extends Arquillian {
      */
     @Test
     public void builderForConcurrencyManagerIsProvided() {
+        ConcurrencyProvider provider = ConcurrencyProvider.instance();
+        ClassLoader classLoader = ConcurrencyManagerTest.class.getClassLoader();
+        Builder concurrencyManagerBuilder = null;
+
         try {
-            ConcurrencyProvider provider = ConcurrencyProvider.instance();
-            ClassLoader classLoader = ConcurrencyManagerTest.class.getClassLoader();
-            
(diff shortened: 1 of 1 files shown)


genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ConcurrencyManagerTest.java
```diff
@@ -0,0 +1,93 @@
...
+            Assert.assertEquals(builtManager, registeredManager,
+                    "ConcurrencyManager.getConcurrencyManager(classLoader) did not return the same manager that was registered.");
+            
+            //release the ConcurrencyManager
+            provider.releaseConcurrencyManager(registeredManager);
+            Assert.assertNotEquals(builtManager, provider.getConcurrencyManager(classLoader),
+                    "ConcurrencyManager was not released from the ConcurencyProvider.");
+            
+        } 
+        catch (UnsupportedOperationException ex) {
```
njr-11: I think the UnsupportedOperationException should be more granularly placed.  For example, an implementation should not be allowed to support registerConcurrencyManager but reject releaseConcurrencyManager.  Instead, an initial check should be make if concurrencyProvider.getConcurrencyManagerBuilder is supported, and that should set expectations for what the other methods are allowed to raise UnsupportedOperationException for.

nmittles:
@FroMage Do you have the cycles to review this PR? I know you have an interest in this portion of the spec.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/ConcurrencyManagerTest.java
nmittles: Updates for UnsupportedOperationException checking have been made in commit 2.

PR #108, opened 2019-01-23
withContextCapture must reject Async methods without executor because it is not backed by an executor

### LINKED ISSUE #107: ThreadContext.withContextCapture must specify that *Async methods are rejected without an executor
As discussed on gitter, ThreadContext.withContextCapture currently places a requirement on the resulting CompletionStage/CompletableFuture for *Async actions to run inline rather than on a new thread, which is due to there not being any executor.  However, this contradicts the JavaDoc for CompletionStage/CompletableFuture, which states that the action is "executed using this stage's default asynchronous execution facility".  It was decided that a more appropriate behavior would be to reject these methods entirely, on the grounds that there is no "default asynchronous execution facility" when created from a ThreadContext.

njr-11:
pull fixes #107

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
@@ -420,10 +420,11 @@ public interface ThreadContext {
      * <p>Returns a new <code>CompletableFuture</code> that is completed by the completion of the
      * specified stage.</p>
      *
-     * <p>The new completable future is backed by a ThreadContext rather than a ManagedExecutor,
-     * and therefore its default asynchronous execution facility, which is supplied by the container,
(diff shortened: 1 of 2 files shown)


manovotn:
We are going to have to alter existing TCKs with this change.
Should we do that as part of this PR (or at least remove the current checks verifying that invocations are done in one thread)?

njr-11:
Yes, the TCK tests that covered the old behavior will need to be switched.  I was thinking to do that under a separate pull attached to both #107 (withContextCapture updates) and #60 (the TCK issue).  First, I wanted to confirm agreement on the updated requirements in the JavaDoc (this pull).  If you would rather I include everything in this pull, I'd be fine with that as well.

njr-11:
I added a commit to update the TCK tests according to the changes as well.  It turns out that it only impacted two test methods.

PR #109, opened 2019-01-24
Add clarification that MP Config is applied statically.

### LINKED ISSUE #106: Clarify whether MP Config is applied statically or dynamically
Currently the MicroProfile Concurrency spec doesn't specify whether ManagedExecutor and ThreadContext instances created by the container support dynamic MicroProfile Config property updates.

I'd like to propose that the spec is updated to clarify that these container managed instances have MP Config properties applied statically, and do not respond to dynamic property updates. I believe there is little to be gained in the required overhead to support such dynamic updates.

Here is a suggested statement we could add:
"For ManagedExecutor and ThreadContext instances produced by the container, MicroProfie Config is applied statically. These instances will not dynamically respond to MicroProfile Config changes made after they are produced."

nmittles:
pull fixes #106 

### DIFF
--- a/spec/src/main/asciidoc/mpconfig.asciidoc
+++ b/spec/src/main/asciidoc/mpconfig.asciidoc
@@ -17,7 +17,7 @@
 [[concurrencympconfig]]
 == Integration with MicroProfile Config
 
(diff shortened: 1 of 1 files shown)


genie-microprofile:

PR #111, opened 2019-01-25
clarify ThreadContext propagation for *Async methods that are supplied with a different executor

### LINKED ISSUE #110: Spec is unclear about context propagation when different executor supplied to Async methods
In addition to the usual patterns,

`stage2 = stage1.thenRun(action2);`

`stage3 = stage2.thenRunAsync(action3);`

the CompletionStage/CompletableFuture API also allow you to override the executor that is used to run the action for a particular stage via methods like,

`stage4 = stage3.thenRunAsync(action4, executor);`

In the former two cases, it is clear that the managed executor's own configuration of propagated/cleared context types determine how thread context is captured and later applied when running the action.

However, in the latter case, the user is explicitly asking for a different executor - one that can be a managed executor as well and might have an entirely different configuration for propagated/cleared context types.  The spec doesn't currently clarify which executor's configuration is used in this case.

We should clarify this so that implementations will be consistent and users will know what behavior to rely upon.

njr-11:
pull fixes #110 

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ManagedExecutor.java
@@ -42,10 +42,11 @@ import org.eclipse.microprofile.concurrent.spi.ConcurrencyProvider;
  *    ...
  * </pre>
  *
- * <p>This specification allows for managed executors that do not capture and propagate thread context,
- * which can offer better performance. If thread context propagation is desired only for specific stages,
- * the <code>ThreadContext.contextual*</code> API methods can be used to propagate thread context to
- * individual actions.</p>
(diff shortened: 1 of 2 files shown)


njr-11:
updated to reflect the majority view and also to include the detail about pre-contextualized actions taking precedence, which had previously been lacking from the JavaDoc

njr-11:
I haven't seen any further feedback and so given that the outcome is now 4 to 2 in favor of the proposal to use the thread context propagation of the ManagedExecutor which is the default asynchronous execution for the stage rather than override it when the supplied executor is also a ManagedExecutor, let's go with the preference of the majority on this.  This pull is updated accordingly.
@FroMage or @manovotn could you review it?
I'll also look into getting a TCK scenario added for this, but as a separate pull.

PR #113, opened 2019-01-29
TCK tests for programmatic CDI lookup of ManagedExecutor and ThreadContext based on NamedInstance qualifier

njr-11:
Write some TCK tests to cover programmatic CDI lookup, verifying that this is possible for ManagedExecutor and ThreadContext based on the NamedInstance qualifier, but that MP Concurrency implementations do not provide for programmatic lookup of unqualified instances

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
@@ -32,6 +32,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
+import javax.enterprise.inject.Instance;
+import javax.enterprise.inject.literal.NamedLiteral;
+import javax.enterprise.inject.spi.CDI;
 import javax.inject.Inject;
 import javax.inject.Named;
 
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/MPConfigTest.java
```diff
@@ -738,4 +743,121 @@ public void overrideMaxQueuedWith4ForManagedExecutorFieldWithConfigAndName()
...
         }
     }
+
+    /**
+     * CDI programmatic lookup of ManagedExecutor without any qualifiers must not find anything.
+     * This test is possible for ManagedExecutor because the test application does not provide any
+     * producers of ManagedExecutor without qualifiers.
+     */
+    @Test(dependsOnMethods = "beanInjected")
+    public void programmaticLookupManagedExecutorWithoutQualifiers() {
```
aguibert: we should have an equivalent test for ThreadContext as well.

Also, I believe it was agreed on that the container would *block* any attempts by the application to register an `@Default` producer. If I'm remembering that correctly, let me know and I"ll add a test scenario for it in the tracking work item.

njr-11: I couldn't add a test for ThreadContext to this test application because it already had a producer for ThreadContext with different qualifier, and that gets returned instead of indicating unsatisfied.  The select method must assume `@Any` when no qualifiers are present.

njr-11: And yes, I think you are remembering correctly about there being a decision to block/fail the application from registering a default producer for ManagedExecutor/ThreadContext, although I can't remember that actually being written down in the spec, so when we get to that, it may need a sentence added to JavaDoc or spec a well if it isn't already there.

PR #114, opened 2019-01-30
TCK tests for cdi injection of ThreadContext

njr-11:
Write TCK tests to cover the following scenarios:
- The container creates a ThreadContext instance per unqualified ThreadContext injection point
- A ThreadContext instance that is injected by the container can be configured via `@ThreadContextConfig`
- If the `@NamedInstance` qualifier is present on a ThreadContext injection point that is annotated with `@ThreadContextConfig`, the container creates a ThreadContext instance that is qualified by the name value

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
@@ -400,6 +400,9 @@ public class CDIBean {
      * Application-defined producer methods can have injection points of ThreadContext that are provided by the container.
      */
     public void testAppDefinedProducerUsingInjectedThreadContext() {
+        Assert.assertNotNull(priority3Executor,
+                "Application should be able to create its own CDI producer that injects a ThreadContext that is provided by the container.");
+
         int originalPriority = Thread.currentThread().getPriority();
         int newPriority = originalPriority == 2 ? 1 : 2;
         try {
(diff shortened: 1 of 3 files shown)


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
```diff
@@ -190,4 +252,251 @@ public void checkMaxQueued3(ManagedExecutor executor) throws ExecutionException,
...
+                Assert.assertEquals(Label.get(), "testInjectThreadContextEmptyConfig-label",
+                        "Thread context type (Lable) was not propagated.");
+                Assert.assertEquals(Thread.currentThread().getPriority(), newPriority,
+                        "Thread context type (ThreadPriority) was not propagated.");
+                return true;
+            });
+
+            Thread.currentThread().setPriority(4);
+            Label.set("testInjectThreadContextNoConfig-new-label");
+            Buffer.set(new StringBuffer("testInjectThreadContextNoConfig-new-buffer"));
```
nmittles: Copy/paste of "NoConfig" instead of using "EmptyConfig"?

njr-11: Good catch. This doesn't effect the correctness of the test (it tests that a previously captured context is used instead), however, it will be much clearer to use consistent names.  Fixing...

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIBean.java
```diff
@@ -190,4 +252,251 @@ public void checkMaxQueued3(ManagedExecutor executor) throws ExecutionException,
...
+        int originalPriority = Thread.currentThread().getPriority();
+        int newPriority = originalPriority == 2 ? 1 : 2;
+        try {
+            Thread.currentThread().setPriority(newPriority);
+            Label.set("testAppDefinedProducerUsingInjectedThreadContext-label");
+            Buffer.set(new StringBuffer("testAppDefinedProducerUsingInjectedThreadContext-buffer"));
+
+            // priority3Executor is an application-produced instance, where its producer method injects a ThreadContext
+            // instance that is provided by the container. The following verifies thread context propagation of that instance,
+            priority3Executor.execute(() -> {
```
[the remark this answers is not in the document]
njr-11: Sure, I'll add that

PR #121, opened 2019-02-08
Make sure the MPConfigTest has beans.xml file.

manovotn:
The test doesn't add `beans.xml` while other TCK tests with CDI do.
This can cause problems with different Arq. adapters such as the one for SE env which we leverage.

It doesn't impact the tests in any way.


FroMage:
LGTM. Will wait for @njr-11 to review before merging.

PR #126, opened 2019-02-18
Tests for CDI context propagation

aguibert:


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIContextTest.java
```diff
@@ -0,0 +1,164 @@
...
+     * has the state propagated to it when ran on the same thread.
+     */
+    @Test
+    public void testCDITCCtxPropagate() throws Exception {
+        ThreadContext defaultTC = ThreadContext.builder().build();
+
+        requestBean.setState("testCDIContextPropagate-STATE2");
+        Callable<String> getState = defaultTC.contextualCallable(() -> {
+            String state = requestBean.getState();
+            System.out.println("testCDIContextPropagate#2 state=" + state);
```
manovotn: Probably testing leftovers? There are multiple occurrences throughout the test, please remove.

manovotn:
@aguibert Checking out these tests, it seems that they all run in different threads. I cannot reproduce what we talked about originally, e.g. the behaviour when you try to perform context propagation (or clearing) on the same thread.

tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIContextTest.java
aguibert: I was intending to leave in place because it can provide some useful debug information, (especially when impls flow System.out through a logger that includes timestamp and thread ID) in the event of a test failure. I can remove them though

aguibert:
> Also, do I get it right that using the `Callable` you are basically running it on the main thread? E.g. simulating the known culprit we've been discussing?

Yes, directly invoking `call()` on the `Callable` for the ThreadContext tests runs the task on the main thread which was the scenario giving me the trouble originally.  The ManagedExecutor tests where we submit the CompletableFuture via `supplyAsync` may or may not run on the same thread.

aguibert:
@manovotn amended the commit to remove the print statements as you requested. AFAIK that was the only remaining review comment on this PR? If yes, I'll go ahead and merge.

PR #129, opened 2019-02-22
Possible compromise solution for transaction context defaults

### LINKED ISSUE #124: Revisit discussion of transaction context propagation
We don't currently propagate transaction context by default. Instead, we clear (suspend) it, which allows tasks/actions to run their own transactions.

This discussion was brought up again on last week's MicroProfile Concurrency call.  This position was questioned and there is a push from at least one member of the spec group (@FroMage) to make transaction context propagation the default.  I'm opening this issue to formally represent the discussion and document its resolution.

I can provide my own position on the matter, which I'll document here.  Those with the opposite view should feel free to append to this comment with the explanation (I can't adequately represent the opposing position, and in any case would be biased),

**Why we shouldn't propagate transaction context by default:**

* _Consistency with EE Concurrency_. In every other way we are fully consistent with EE Concurrency, which MicroProfile Concurrency is modeled after. EE Concurrency requires that transaction context be suspended when running tasks (equivalent to our `cleared` behavior) which lets the task run its own transactions, while providing a way for applications to programmatically ask for any transaction that is already on the thread to be left there (equivalent to our `unchanged` behavior). Switching our default for transaction context to `propagated` would undermine compatibility and customer expectations.

* _Data integrity_. There are all sorts of ways to go wrong with propagated transaction context, including the possibility that you end up running operations under a transaction that you didn't intend to, or causing premature commit/rollback. Transaction context propagation is dangerous if you don't know exactly what you are doing. It is for advanced usage patterns, so it shouldn't need to be default.

* _Limited applicability_. Many transaction manager and transactional resource implementations won't be able to propagate transaction context all (the very idea is scary to our own transaction manager team and they have strongly warned against doing it). Others will only be able to support it for XA-capable resources and not for one-phase resources, and even then, some types of XA-capable resources will support it whereas others will not. Even when it's the case where the XA-capable resources involved and the transaction manager do support transaction context propagation, some of them will end up limiting support to operations being performed in serial, rather than allowing the transaction to actually be used in parallel, even though the CompletionStage model certainly allows for and encourages parallel use.

* _Defaults ought to reflect core usage patterns_. By enabling transaction context propagation by default, applications that don't have any interest in propagating transactions but happen to make use of completion stages for non-transactional operations within a transaction will be unable to do so, because the transaction manager will need to reject propagation of the transaction. The application will then need to go out of its way to turn off transaction context propagation in order to work properly.

Propagation of transactions is an advanced pattern. To use it, you need to be aware of the various limitations and capabilities of the involved resources and the transaction manager. You need to be willing to read about and follow best practices for transaction propagation just to write safe code. If you are willing to do these things, you should be willing to write code that explicitly asks for transaction context propagation to be turned on specifically where you need it. In fact, I expect that such users will _prefer_ to be explicit about when they want to use transaction context propagation, and they actually wouldn't want it enabled by default across all usage of completion stages within the application because that would add both unnecessary risk and extra overhead.

**The argument in favor of propagating transaction context by default:**

(Please add here)

**If no agreement can be reached**

If no agreement can be reached then a possible fallback plan, which sacrifices usability of the CDI config annotations is the following:

Current defaults (between both ThreadContextConfig and ManagedExecutorConfig),

```
String[] cleared() default { ThreadContext.TRANSACTION };
String[] propagated() default { ThreadContext.ALL_REMAINING };
...
int maxAsync() default -1;
int maxQueued() default -1;
```

Would be changed to,

```
String[] cleared() default { ThreadContext.DEFAULT_CLEARED };       // means TRANSACTION unless MP Config says otherwise
String[] propagated() default { ThreadContext.DEFAULT_PROPAGATED }; // means ALL_REMAINING unless MP Config says otherwise
String[] unchanged() default { ThreadContext.DEFAULT_UNCHANGED };   // means empty set unless MP Config says otherwise
...
int maxAsync() default ManagedExecutor.DEFAULT_MAX_ASYNC;   // means effectively unlimited unless MP Config says otherwise
int maxQueued() default ManagedExecutor.DEFAULT_MAX_QUEUED; // means effectively unlimited unless MP Config says otherwise
```

After having done the above, we would make it possible for the user to override via MicroProfile config as:

```
ManagedExecutor/cleared=
ThreadContext/cleared=
```

The above would override the default for both the config annotations as well as the builders.

njr-11:
Update spec per compromise that transaction context propagation is default for ThreadContext but remains cleared for ManagedExecutor.

fixes #124 if the compromise solution is decided upon.  At this point, discussion is still ongoing (still hoping that others will come to the conclusion that the current spec behavior of default configuration being to clear transaction context is best for ThreadContext as well as ManagedExecutor).

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
@@ -109,11 +109,15 @@ public interface ThreadContext {
          * <p>This set replaces the <code>cleared</code> set that was
          * previously specified on the builder instance, if any.</p>
          *
-         * <p>The default set of cleared thread context types is
-         * {@link ThreadContext#TRANSACTION}, which means that a transaction
+         * <p>For example, if the user specifies {@link ThreadContext#TRANSACTION}
+         * in this set, then a transaction
          * is not active on the thread when the action or task runs, such
          * that each action or task is able to independently start and end
          * its own transactional work.</p>
+         *
+         * <p>By default, the set of cleared context is empty, but context clearing
+         * also depends on whether all context types are accounted for in the
+         * {@link #propagated} and {@link #unchanged} sets.
          *
          * <p>{@link ThreadContext#ALL_REMAINING} is automatically appended to the
          * set of cleared context if neither the {@link #propagated} set nor the
@@ -141,10 +145,8 @@ public interface ThreadContext {
          * <p>The default set of propagated thread context types is
          * {@link ThreadContext#ALL_REMAINING}, which includes all available
          * thread context types that support capture and propagation to other
-         * threads, except for those that are explicitly {@link cleared},
-         * which, by default is {@link ThreadContext#TRANSACTION} context,
-         * in which case is suspended from the thread that runs the action or
-         * task.</p>
+         * threads, except for those that are explicitly {@link #cleared}
+         * or {@link #unchanged}.</p>
          *
          * <p>Constants for specifying some of the core context types are provided
          * on {@link ThreadContext}. Other thread context types must be defined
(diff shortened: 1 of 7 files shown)


njr-11:
Per recent comments in #124 this idea for a compromise is not valid and has been found unacceptable by our lead architects.  Closing/cancelling this pull.

Issue #116, opened 2019-02-04
Support "Checked Transaction" behaviour


mmusgrov:
If there is an active transaction on multiple asynchronous completion stages then it is desirable that we avoid the situation where one thread ends a transaction whilst other threads are still using it (otherwise the work done by the other threads will not be transactional). I think it would help developers write more robust code if we provide a mechanism for the application to deal with the circumstance by allowing the creator to to provide a "Checked Transaction" handler with each managed execution similar to

```java
     public void check(boolean isCommit, Uid txId, List<Long> activeThreads);
```
which is called if a transaction is ended whilst other threads are still using it. The application would then be able to ensure that the business logic being run by threads in the activeThreads list safely terminates.

njr-11:
It seems to me like the most natural way for applications to coordinate asynchronous transaction logic in this programming model would be to use the CompletionStage API itself. For example, using a JTA transaction (oversimplified in that try/catch blocks are omitted),

```
tx.begin()
stage1 = executor.runAsync(action1);
stage2 = executor.runAsync(action2);
stage3 = stage1.runAfterBoth(stage2, () -> tx.commit())
               .exceptionally(failure -> tx.rollback());
```

However, having written this, what concerns me here is that the main thread still has a transaction hanging around.  I think if this pattern is going to naturally fit with completion stages, we would need some sort of suspend operation from the transaction in order for the user to take it off of the main thread so that they can proceed to do other work while the transaction continues asynchronously.

FroMage:
Also note that async actions are likely to not have thread IDs at all, since they're more often tasks that have no specific thread.

mmusgrov:
So I think we just need to document the recommended best practice:
- for running transaction code in async stages,
- that the main thread should suspend the transaction, and
- that transaction termination should run in an after stage?

njr-11:
Yes, documentation of this as the best practice, and probably including a simple example, should be all we need.

mmusgrov:
Okay, I will do that and propose it here, thanks.

FroMage:
So here's an example of how containers have to be adapted to support async propagation:

```java
HttpServletRequest req = ...;
try {
    requestContext.activate();
    return action.call(exchange, context);
} finally {
    if(req.isAsyncStarted()) {
        requestContext.deactivate();
        req.getAsyncContext().addListener(new AsyncListener() {
            @Override
            public void onComplete(AsyncEvent event) throws IOException {
                requestContext.terminate();
            }

[...]

PR #132, opened 2019-02-28
Issue116: Document how to manage transaction propagation

mmusgrov:
#116 

This PR adds an example to the documentation of how to manage transaction propagation.

### DIFF
--- a/spec/src/main/asciidoc/examples.asciidoc
+++ b/spec/src/main/asciidoc/examples.asciidoc
@@ -65,7 +65,11 @@ This section includes some additional examples of spec usage.
     });
 ----
 
-=== Run under the transaction of the executing thread
+=== Transaction Usage Guidelines
+
+For implementations that support propagation of JTA transactions, the following guidelines and examples demonstrate possible common usage patterns:
+
+==== Run under the transaction of the executing thread
 
 [source, java]
 ----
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
I disagree that MP concurrency should contain any transaction test. As said yesterday, MP Concurrency should define a framework on how to opt in for context propagation not defining what should be propagated.

njr-11:
> I disagree that MP concurrency should contain any transaction test.

This comment is out of place under this pull.  This pull does not deliver any tests.  It documents and example code snippet of using the MP Concurrency spec in conjunction with a JTA implementation that propagates context.  We have other examples snippets that illustrate other context types, such as propagation of the java:comp/env namespace so that you can look up a data source, so I cannot see why we would omit transaction context examples.  The fact that we gave an example of JNDI doesn't mean that all implementation must support JNDI, nor does including an example of transaction context propagation mean that all implementations must support JTA or transaction context propagation.  I could see prefacing these sort of examples with something along the lines of, "For implementations that support propagation of JTA transactions, the following guidelines and examples demonstrate possible common usage patterns, ..." but it would take away from the usability of the spec not to have good examples that users can refer to.

mmusgrov:
I added a commit that points out that transaction propagation is optional.

njr-11:
Cancelling this as discussed under 140.

Issue #16, opened 2018-09-18
Question: CDI request context usage


mkouba:
CDI request context is by its nature single-threaded, i.e. users never synchronize `@RequestScoped` beans. What happens if a user sumbits a task to the executor and then continues to work with the current request context:

```java
class MyService {
  // RequestBean is @RequestScoped
  @Inject 
  RequestBean bean;

  @Inject 
  ManagedExecutor executor;
  
  @Inject 
  ThreadContext threadContext;
  
  void doSomething() {
    // myFunction might be executed in parallel
    executor
      .supplyAsync(threadContext.withCurrentContext(myFunction)); 
     // what happens if change the current request context?
     bean.changeMyState();
  } 
```
Without synchronization (on the bean class level) race conditions may occur.

FroMage:
My personal opinion is that this should be up to the user to synchronise concurrent modifications of state across threads. We can only guarantee that contexts are propagated. It is then up to the user to use the proper thread synchronisation or the proper underlying thread model (in vert.x, each request is bound to a single thread).

FroMage:
Well, ATM context is not propagated so it's even easier to break an app now, even without any concurrent access going on ;) It would still be progress and less breakage.

mkouba:
Well, that's probably a different kind of "breaking" ;-). Anyway, I do not question the usefulness of context propagation. I'm trying to point out the difficulties before it's too late...

njr-11:
I think there would be some value in updating the spec doc to emphasize the application's responsibility for using each type of context in a way that is thread safe.  Generally this is something that each provided context type will need to define itself.  We could also consider whether CDI context ought to be listed as one of the types that is propagated by default when the user injects/builds a ManagedExecutor or ThreadContext without specifying a particular set of context types.  And let's get some input from CDI experts such as @Emily-Jiang around whether there is anything more that ought to be done here particular to CDI.

Emily-Jiang:
@mkouba is right. This is indeed an issue. We should discuss all context propogation further in our next hangout.

manovotn:
@njr-11 @Emily-Jiang Since I cannot really make in into those mtg (due to their time), have you came up with anything regarding this question?

What @mkouba brought up is IMO a very good and serious remark. 

[...]

PR #138, opened 2019-03-07
Clarify javadoc for ThreadContext.CDI

aguibert:
For users that are not intimately familiar with CDI, it may not be clear what CDI context propagation means or why the user should care.

Add a code example to the ThreadContext.CDI javadoc to illustrate the value of propagating CDI context.


aguibert:
Fixes #16

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
```diff
@@ -240,6 +240,44 @@ public static Builder builder() {
...
+            // do some work with the request state
+        }).thenApply(more -> {
+            // request state also available in future stages
+        });
+    }
+    </pre></code>
+     *
+     * If CDI context is 'unchanged', access to CDI bean's contextual state 
+     * will be non-deterministic. Namely, context may be missing, or context
+     * from a different
```
njr-11: It looks like this sentence is prematurely cut off here.

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
```diff
@@ -240,6 +240,44 @@ public static Builder builder() {
...
+     * <code><pre>
+    &#64;RequestScoped
+    public class MyRequestScopedBean {
+        public String getState() {
+          // returns some request-specific information to the caller
+        }
+    }
+    </pre></code>
+     * 
+     * When CDI context is propagated, the request's state will also
```
manovotn: I would change it to something like:

_CDI context propagation includes request, session and conversation contexts.
When CDI context is propagated, all of the above mentioned contexts that are currently active will be available to the contextualized task with preserved state._

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
```diff
@@ -240,6 +240,44 @@ public static Builder builder() {
...
+        exec.supplyAsync(() -> {
+            String state = reqBean.getState();
+            // do some work with the request state
+        }).thenApply(more -> {
+            // request state also available in future stages
+        });
+    }
+    </pre></code>
+     *
+     * If CDI context is 'cleared', the state of any Request, Session, or Conversation
```
manovotn: I'd drop the notion of restoring contexts, that's implied by the nature of propagation (as is cleanup after you propagate it). Furthermore, we need to stress out that `cleared` means you will have the same contexts active but with no state it them.

_If CDI context is `cleared`, currently active contexts will still be available to the contextualized task, but their state will be erased._

api/src/main/java/org/eclipse/microprofile/concurrent/ThreadContext.java
```diff
@@ -240,6 +240,44 @@ public static Builder builder() {
...
+     * scoped beans will be cleared during task execution, and restored after execution.
+     * 
+     * If CDI context is 'unchanged', access to CDI bean's contextual state
```
manovotn: Hmm, this is somewhat tricky. I'd almost lean towards saying it's bad practice to leave it `unchanged` or otherwise discourage users from going down this path _if they want to use any CDI beans_.
Because based on what thread runs it, you may be able to access beans (if on main thread) or you will hit a wall head on with `ContextNotActiveException` (on any new thread).

aguibert:
thanks for the review @manovotn, I've added another commit to address your review comments

PR #139, opened 2019-03-08
remove config annotations and NamedInstance

### LINKED ISSUE #135: @ThreadContextConfig / @ManagedExecutorConfig are superfluous
Currently we have two different ways for a user to configure a `ThreadContext` or a `ManagedExecutor`:

- in procedural code using a `Builder`, or
- using a `Config` annotation that performs injection.

The annotation-based approach is ever so slightly less verbose, and the builder-based approach lets one specify attributes based on runtime conditions, but other than that there's not much difference AFAICT.

It's not necessary to have two ways to do this thing. Assuming that the extra flexibility of a `Builder` is actually truly useful, then I would get rid of the annotation. It is *very* easy in CDI to create a little producer method that builds something and binds it to the CDI application scope. It is *not* the CDI way to have injection annotations with lots of configuration attributes that then get copy/pasted all over the codebase.

njr-11:
fixes #135

This pull request removes the annotations ManagedExecutorConfig, ThreadContextConfig, and NamedInstance, including all tests cases and other references to them through the spec.  In several places, I've updated spec doc examples or tests to use the builders rather than these removed annotations, with the aim of preserving non-duplicated test coverage where possible.  The CDI section of the spec doc is replaced with best practices examples of defining producers as ApplicationScoped.


spec/src/main/asciidoc/overview.asciidoc
```diff
@@ -68,101 +68,69 @@ Instances of `ManagedExecutor` and `ThreadContext` can be constructed via builde
...
 ----
 
-=== Sharing Configured Instances
+=== Sharing Instances
 
-Configured instances of `ManagedExecutor` and `ThreadContext` are shared based on their unique name, which is assigned by the CDI qualifier, `@NamedInstance`.
-When one injection point is annotated with both a configuration and a `@NamedInstance` qualifier, the application can inject the same instance elsewhere by annotating other injection points with just the matching qualifier.
-In the example below, the `executor1` injection point defines a configuration and a name `myExec` for a configured `ManagerExecutor` instance.
-The other injection points, `executor2` and `executor3`, share the same `ManagedExecutor` instance by using the `@NamedInstance` qualifier with the name `myExec`.
+CDI application scope is a convenient mechanism for sharing instances across an application.
```
manovotn: `CDI producers combined with injection are a convenient mechanism for sharing...`

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -43,22 +43,12 @@ To get started, add this dependency to your project:
...
-----
-
-Or you can create one using a builder:
+Use the fluent builder API to construct a `ManagedExecutor`:
 
 [source,java]
 ----
     ManagedExecutor executor = ManagedExecutor.builder()
                        .propagated(ThreadContext.APPLICATION, ThreadContext.CDI)
-                       .cleared(ThreadContext.ALL_REMAINING)
```
manovotn: Accidental removal?

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -71,3 +61,23 @@ Then obtain a `CompletableFuture` or `CompletionStage` from the `ManagedExecutor
                                              .thenApplyAsync(function1)
                                              .thenApply(function2);
 ----
+
+Take care to shut down managed executors once the application no longer needs them:
```
manovotn: Do you think it might be worth to show the disposer method as well? Not necessarily in this file.

api/src/main/java/org/eclipse/microprofile/concurrent/spi/ThreadContextProvider.java
```diff
@@ -107,8 +107,7 @@
      *
      * <p>For example:</p>
      * <pre><code>
-     * &commat;Inject &commat;ManagedExecutorConfig(propagated = ThreadContext.CDI)
-     * ManagedExecutor executor;
+     * ManagedExecutor executor = ManagedExecutor.builder().propagated(ThreadContext.CDI).build();
```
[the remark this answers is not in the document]
njr-11: Good catch, I'll get it added.

spec/src/main/asciidoc/builders.asciidoc
```diff
@@ -60,6 +60,13 @@ It is the responsibility of the application to shut down `ManagedExecutor`
 instances that are no longer needed, so as to allow the container to
 efficiently free up resources.
 
+The MicroProfile Concurrency specification does not require `ManagedExecutor`
```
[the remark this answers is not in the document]
njr-11: sorry, I misremembered from the last call, thinking you had asked it to be mentioned in the spec.  I'll remove this section.  I agree that the spec does not need to mention this either way.

spec/src/main/asciidoc/cdi.asciidoc
```diff
@@ -17,95 +17,120 @@
...
-Injection points with the `@NamedInstance` qualifier have a name that is explicitly declared by the qualifier.
-Injection points that have no CDI qualifier on them have a unique name which is deduced from the injection point itself.
-Container provides, per unique name, an instance of either `ManagedExecutor` and/or `ThreadContext` unless the application specifies a CDI producer with `@NamedInstance` qualifier, given unique name and respective type.
+When `ManagedExecutor` and `ThreadContext` are used as CDI beans, the best practice is to define their producers as `@ApplicationScoped` so that instances are shared and reused. In most cases, more granular scopes are undesirable, such as having a new `ManagedExecutor` instance per request. In the event that a more granular scope is desired, the application must take care to supply a disposer to ensure that the executor is shut down once it is no longer needed. When using application scope, it is optional to supply a disposer because the specification requires the container to automatically shut down `ManagedExecutor` instances when the application stops.
```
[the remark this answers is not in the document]
njr-11: Thanks for the improvements here.  I'm updating it as you suggested.

spec/src/main/asciidoc/overview.asciidoc
njr-11: Updating as recommended.

spec/src/main/asciidoc/release_notes.asciidoc
njr-11: Sure, I can add an example disposer

spec/src/main/asciidoc/release_notes.asciidoc
njr-11: Yes, good catch. I'll get this line added back.

README.adoc
```diff
@@ -108,13 +108,19 @@ currently as well as after adoption into Jakarta EE.
 Injection
 ^^^^^^^^^
 
-It shall be possible to inject instances of ManagedExecutor and
-ThreadContext via the `@Inject` annotation, for example:
+It shall be possible to build instances of ManagedExecutor and
```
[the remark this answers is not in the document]
njr-11: I had intended to completely switch this section over to builders (and forgot to update the title) given that it is was previously an example of the injection approach that was being removed.
That's a good idea to also given an example of the producer approach here, so I'll add that.

njr-11:
@manovotn I've added a commit with fixes to all of the code review issues

PR #140, opened 2019-03-12
document exceptions raised when optional transaction context is not supported

### LINKED ISSUE #137: Document options for more granular transaction context support
Define exception type (possibly IllegalStateException) and where it can be used

njr-11:
Add documentation of where exceptions can be raised and which exceptions to raise in various places where transaction context propagation (which is optional) can be rejected.  Update TCK tests to allow for these exceptions.

fixes #137

### DIFF
--- a/spec/src/main/asciidoc/txcontext.asciidoc
+++ b/spec/src/main/asciidoc/txcontext.asciidoc
@@ -166,7 +166,9 @@ try {
 
 === Propagation of Active Transactions for Parallel Use
 
-An implementation that supports the optional behavior of propagating active transactions for use on multiple threads in parallel does not raise `IllegalStateException` at any point. As always, the application is responsible for following best practices to ensure transactions are properly resolved and transactional resources are properly cleaned up under all possible outcomes.
+An implementation that supports the optional behavior of propagating active transactions for use on multiple threads in parallel may choose whether or not to support commit and rollback operations from dependent stage actions. If unsupported, these operations raise `SystemException` when invoked from a separate completion stage action. As always, the application is responsible for following best practices to ensure transactions are properly resolved and transactional resources are properly cleaned up under all possible outcomes.
+
+Here is an example of committing the transaction in a dependent stage action,
 
 [source, java]
 ----
@@ -178,9 +180,9 @@ try {
                    .whenComplete((result, failure) -> {
         try {
             if (failure == null && tx.getStatus() == Status.STATUS_ACTIVE)
-                tx.commit();
+                tx.commit();   // <-- raises SystemException if unsupported within dependent stage
             else
-                tx.rollback();
+                tx.rollback(); // <-- raises SystemException if unsupported within dependent stage
         } catch (Exception x) {
             if (failure == null)
                 throw new CompletionException(x);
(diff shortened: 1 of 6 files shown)


spec/src/main/asciidoc/txcontext.asciidoc
```diff
@@ -0,0 +1,193 @@
...
+    }).thenApply(updateCount -> {
+        try (Connection con = dataSource.getConnection()) {
+            return updateCount + con.createStatement().executeUpdate(sql3);
+        } catch (SQLException x) {
+            throw new CompletionException(x);
+        }
+    }).whenComplete((result, failure) -> {
+        try {
+            if (failure == null && tx.getStatus() == Status.STATUS_ACTIVE)
+                tx.commit();
```
mmusgrov: So this executor thread has finished the transaction which is still associated with the thread that called tx_begin on line 102. This means that that thread (the one that created the transaction) is now associated with an invalid transaction.

In the PR #132 I documented best practice I added the following warning:
{quote}
The issue with ending the transaction in a completion stage is that
the context of the initiating thread will be re-applied after the executor finishes and if a stage
ended the transaction then this initiating thread will be associated with the finished transaction
which would mean that that thread could not begin any new transactions.
{quote}

spec/src/main/asciidoc/txcontext.asciidoc
```diff
@@ -0,0 +1,193 @@
...
+----
+tx.begin();
+try {
+    stage1 = executor.runAsync(action1);
+    stage2 = executor.runAsync(action2);
+    stage3 = stage1.runAfterBoth(stage2, (u,v) -> action3)
+                   .whenComplete((result, failure) -> {
+        try {
+            if (failure == null && tx.getStatus() == Status.STATUS_ACTIVE)
+                tx.commit();
```
mmusgrov: Same comment as above

spec/src/main/asciidoc/txcontext.asciidoc
njr-11: That's a good point - it looks like I've missed out on maintaining consistency with 132 in this area.
Although I do wonder if such a limitation is going to be too great of an impediment to anyone wanting to use transaction context propagation in reactive patterns.  And also, we don't seem to have that limitation in our own implementation of transaction context propagation (which is all experimental at this point).  I remember that I did need to modify a few lines of code to check that the transaction is still active before restoring it the thread, but we've never had problems with subsequent transactions being disallowed on the thread after that.

From the user's standpoint, it certainly seems the natural thing to let the chain of dependent stages resolve the transaction.  I wonder if this needs to be another of the optional patterns for transaction context propagation.

njr-11: I just added a commit that explains the above as another optional behavior, along with the exception to expect if unsupported, and includes an example of committing from the main thread instead.  Let me know if this works.  Also, it might make sense to put the changes from 132 into txcontext.asciidoc rather than the general examples asciidoc now that we have a good place for it.  I'm really sorry that pull has been held up so long.

mmusgrov: Okay, I think you are correct. I will update our SmallRye implementation to perform the same check that you are making, namely to check that the transaction is still active before deciding whether or not to restore it to the thread, thanks. I will also need to update the guidelines I wrote on the PR 132 (for issue #116).

mmusgrov: @njr-11 In fact I think that the doc updates in this PR are more complete than mine so I would not object if you were to close my PR.

njr-11: I'll go ahead and close it out for now. Feel free to reopen if there ends up being anything else to add.  Thanks for all of your collaboration in this area. It has been very helpful.

PR #143, opened 2019-03-20
Correct some TCKs to fully define TC now that there are no defaults.

manovotn:
Found these three tests that do not define all aspects of `ThreadContext` which is wrong now that we do not have any defaults.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIContextTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIContextTest.java
@@ -135,6 +135,7 @@ public class CDIContextTest extends Arquillian {
         ThreadContext defaultTC = ThreadContext.builder()
                                                .propagated(ThreadContext.CDI)
                                                .cleared(ThreadContext.ALL_REMAINING)
+                                               .unchanged()
                                                .build();
 
         requestBean.setState("testCDIContextPropagate-STATE2");
@@ -154,6 +155,7 @@ public class CDIContextTest extends Arquillian {
         ThreadContext clearAllCtx = ThreadContext.builder()
                         .propagated() // propagate nothing
                         .cleared(ThreadContext.ALL_REMAINING)
+                        .unchanged()
                         .build();
 
         requestBean.setState("testCDIThreadCtxClear-STATE1");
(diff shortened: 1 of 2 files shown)


tck/src/main/java/org/eclipse/microprofile/concurrency/tck/cdi/CDIContextTest.java
```diff
@@ -135,6 +135,7 @@ public void testCDITCCtxPropagate() throws Exception {
         ThreadContext defaultTC = ThreadContext.builder()
                                                .propagated(ThreadContext.CDI)
                                                .cleared(ThreadContext.ALL_REMAINING)
+                                               .unchanged()
```
njr-11: This brings up a good point.  Under the issue/pull to remove defaults, I did so only for propagated/cleared, given that the dispute was over the defaults for those categories.  So I didn't add `unchanged()` to test cases because I was under the impression that the default for `unchanged` remained an empty set.  However, I just looked over the change history and it turns out that the spec never did specify a default for `unchanged` -- we were all just assuming it to be the empty set, or more likely had simply overlooked specifying it in the JavaDoc.

So now, this brings up the question: If we are all still in agreement that empty makes sense as the default for `unchanged`, should we add that language to the JavaDoc in order to spare users from having to specify it?

manovotn: That boils down to how we handle unlisted contexts. And from spec wording, those are currently `cleared`. So we now _assume_ that `unchanged` is empty just like you say. But that is simply an assumption made by implementations. And in a state where we have removed defaults, it seems somewhat inconsistent to have `unchanged` defined with default value. Especially since you know that impls will have their own defaults anyway (not that we would override this one) which can mean this one could effectively be ignored?

We can discuss this tomorrow during meeting if you want(should I create an issue?), but either way, I think these TCK changed should go in so that we leave no room for ambiguity.

njr-11: I agree with this change going in regardless of the decision, because regardless of whether we specify a default for `unchanged`, your changes remain valid.

Here's the case for why a default for `unchanged` would add value and improve the user experience:
We still have language outlining where ALL_REMAINING gets placed (appended to `cleared`) if not specified anywhere else.  Thus, if we had a spec default of empty for `unchanged`, and the vendor defines some sort of default for `cleared` (even empty), to which ALL_REMAINING can be appended, then the user would only need to configure the `propagated` attribute in order to validly configure a ThreadContext instance, just as they would only need to specify `propagated` in order to validly configure a ManagedExecutor.  This is nice because app developers using context propagation will typically think in terms of what context they want/need rather than the context types they don't know of or care about. And, they will only need to configure a single attribute, instead of multiple.

PR #146, opened 2019-03-23
default value for unchanged context types missing from JavaDoc

### LINKED ISSUE #144: JavaDoc for unchanged context types is missing a default value
ThreadContext JavaDoc for the `unchanged` method is missing its default value.  This was discussed under #143 and is being covered under this issue, which we went over on the latest MP Concurrency call, with all in agreement.   Formalizing the default of empty set for `unchanged`, in combination with the already-present spec language that appends `ALL_REMAINING` to the `cleared` context list in the event that `ALL_REMAINING` isn't specified anywhere else, makes possible an optimization where the user only needs to configure the `propagated` context types.  This make configuration simpler for users and provide better usability.

Note: in order to avoid massive merge conflicts, we will plan on postponing this change until right after the spec rename, which, due to changing packages, touches and moves basically every file in the spec.

njr-11:
Pull fixes #144

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/context/ThreadContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/context/ThreadContext.java
@@ -186,7 +186,9 @@ public interface ThreadContext {
          * <p>The MicroProfile Config property, <code>ThreadContext/unchanged</code>,
          * establishes a default that is used if no value is otherwise specified.
          * The value of the MicroProfile Config property can be the empty string
-         * or a comma separated list of context type constant values.</p>
+         * or a comma separated list of context type constant values. If a default
+         * value is not specified by MicroProfile Config, then the default value
+         * is an empty set.</p>
          *
          * <p>The configuration of <code>unchanged</code> context is provided for
          * advanced patterns where it is desirable to leave certain context types
--- a/spec/src/main/asciidoc/overview.asciidoc
+++ b/spec/src/main/asciidoc/overview.asciidoc
@@ -61,6 +61,7 @@ Instances of `ManagedExecutor` and `ThreadContext` can be constructed via builde
 
     ThreadContext threadContext = ThreadContext.builder()
         .propagated(ThreadContext.APPLICATION, ThreadContext.CDI)
+        .cleared(ThreadContext.ALL_REMAINING)
         .build();
 ----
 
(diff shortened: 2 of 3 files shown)


njr-11:
@manovotn could you review this?  It has the changes we discussed during the previous call and during the review of #143

manovotn:
Oh sorry, slipped my attention.
I am still not sure this is wise move. It sort of contradicts the implementations being able to set their own defaults because this one is set in stone for all impls. I know it doesn't contradict yours or our impl but still...
ATM we declare where "Remaining" is (in cleared), but that's quite different from setting a default to some value, isn't it?

njr-11:
It works because, unless there is a conflict between any vendor-defined defaults for `cleared` and what you specify for `propagated` (which is already an error path), then the set of vendor-specific defaults for `cleared` union with ALL_REMAINING will always have the same meaning as ALL_REMAINING when `unchanged` is the empty set (or similar if the user sets it to something).

I may have misinterpreted the resolution and though that you were in favor of this proposal.  But as I now understand it, you wanted only the default value, not any explicit codification of what is possible because of it.  I just took that part out in a latest commit.

manovotn:
No, what I mean is there might be an implementation that would favor not to care about, say, `Transaction` context and would like to put it into `unchanged` by default. This change will prohibit that approach.

I am not sure such approach makes much sense in the first place though. Just thinking aloud...

njr-11:
The idea that any context type would default to `unchanged` seems contrary to the purpose of the specification, which is to make thread context deterministic and reliable when using completion stages.  Really the `unchanged` value is only there for advanced patterns where the user is able to guarantee they have more complete control over the conditions under which a contextualized action/task can run.  Defining consistent behavior is more important than allowing for the possibility that some other vendor -- yet to be identified -- might unilaterally decide that a certain context type doesn't matter (in which case why did they even supply a ThreadContextProvider for it?).  This is why we originally planned to define defaults for all of the categories: `unchanged`, `cleared`, and `propagated`.  When the decision was made to remove defaults for `cleared` and `propagated`, it was not because we wanted to allow flexibility to implementation to do non-standard things, but rather because we could not agree on what those defaults ought to be.  We did that in the hope that we would be able to agree in a vNext.  If we cannot agree on `unchanged` either, then I'll cancel this pull, and just go ahead with the release candidate.  Otherwise, I've been delaying that to allow this change in, if it is wanted.

PR #155, opened 2019-04-02
Remove extraneous <url> tag from the TCK's pom.xml.

nmittles:
Removing an extraneous <url> tag from the TCK's pom.xml that was causing maven.apache.org to show up as the HomePage for the project on maven central.


genie-microprofile:

PR #157, opened 2019-04-15
 mark transaction for rollback only after rejection of parallel use

### LINKED ISSUE #156: TCK test expects transaction commit after rejection of parallel use
We just tried to update our implementation to mark transactions for rollback only after rejecting propagation of a transaction to multiple threads at once, and ran into an error in a TCK test that is erroneously expecting to commit a transaction after the rejected parallel use.  All other TCK tests expect to do a rollback in this case.  The test case needs to be fixed, and I'd also like to add some language to the spec recommending that a transaction ought to be marked as "rollback only" when parallel usage of the transaction is rejected.  It seems wrong to allow such a transaction to commit when some of the operations that should have been performed within it were actually rejected.  This will help the user get a consistent behavior because they will not be led to believe that their commit succeeded when some of the stages with transactional operations didn't run at all.

njr-11:
pull fixes #156

### DIFF
--- a/spec/src/main/asciidoc/txcontext.asciidoc
+++ b/spec/src/main/asciidoc/txcontext.asciidoc
@@ -133,7 +133,7 @@ try {
 stage1.complete(initialCount);
 ----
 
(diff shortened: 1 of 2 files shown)


njr-11:
> Changes look good but was there a particular reason why the tests don't always call rollback for the finally step.
No reason that I know of.   It would certainly be valid to do that, and a bit simpler because we wouldn't need it to track whether or not the attempt to run in parallel was rejected. Would you prefer that I change it to that?

mmusgrov:
No need to make any changes to the PR, it was just a query and you have answered it, thanks.

PR #169, opened 2019-07-15
Added link to spec, removed contents that is in spec and out of date

### LINKED ISSUE #161: Publish the spec and javadoc
We should put the (rendered) spec and javadoc online _somewhere_ because ATM it's impossible to link to them.

Is there any formal placement for those @kenfinnigan ?

FroMage:
This could be enough to fix #161.

I added a badge to the rendered spec, and removed all the outdated content that is in the spec.

Do you guys think we need more text in the README?


README.adoc
```diff
@@ -18,171 +18,14 @@
...
 
 [[microprofile-context-propagation]]
 = MicroProfile Context Propagation
 
 :toc:
 
-* Status: *new*
-* Decision Notes:
-https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/microprofile/jKFu-IS_U90[Discussion
-thread topic with background and covering the design]
```
njr-11: Thanks for catching the out-of-date material and removing it.  I agree with removing the above, however, removing what follows after this would leave us with an empty doc on the main page, so I think it would be better to just update more specific out-of-date wording.  I'll add some review comments below attempting to do that.

README.adoc
```diff
@@ -18,171 +18,14 @@
...
-
-[[introduction]]
-Introduction
-~~~~~~~~~~~~
-
-The proposal introduces APIs for obtaining CompletableFutures that are
```
njr-11: can change "The proposal" to "This specification"

README.adoc
```diff
@@ -18,171 +18,14 @@
...
-It is intended that ThreadContext methods will one day be added to
-EE Concurrency's ContextService and for a single implementation to
-be capable of simultaneously implementing both interfaces, both
-currently as well as after adoption into Jakarta EE.
-
-[[builders]]
-Builders
-^^^^^^^^^
-
-It shall be possible to build instances of ManagedExecutor and
```
njr-11: "It shall be possible to ..." can change to, "With MicroProfile Context Propagation, you can ..."

README.adoc
FroMage: I thought that initially too, except we do have a link to the rendered spec so it's easy to get to the info, and whatever we duplicate here will always end up out-of-date.

I think this file should concentrate on things that are not in the spec.

A short intro of a single paragraph should be fine, though.

PR #171, opened 2019-08-14
CDIContextTest should allow for inactive contexts, split the tests.

### LINKED ISSUE #170: Make propagation of @ConversationScoped optional for MP CP impls
Currently, the spec (with regards to CDI propagation) requires to support request/session/conversation context. It's not really in the `/spec` but I found a notion of that in `ThreadContext` javadoc. Link [here](https://github.com/eclipse/microprofile-context-propagation/blob/master/api/src/main/java/org/eclipse/microprofile/context/ThreadContext.java#L266).

There is also [the CDI test in TCK suite](https://github.com/eclipse/microprofile-context-propagation/blob/master/tck/src/main/java/org/eclipse/microprofile/context/tck/cdi/CDIContextTest.java), that is all-in-one for all scopes and requires conversation scope as well as the others.

I suggest to change this up and make `@ConversationScoped` support optional. The scope is pretty old and not so commonly used as the other two. In fact I would argue this is useful on bigger server side application (possibly clustered and all that) whereas in many other microprofile apps this doesn't really have a spot (request scope is by far the most used one).
For instance in Quarkus (more accurately in Arc, it's CDI impl), there isn't even conversation scope at all and we have seen no requests for it so far. So, just like we support transaction propagation optionally, I think we should do the same for conversation scope.

Alternatively, we could change the TCK tests to first check if a context is active via [`BeanManager.getContext()`](https://github.com/eclipse-ee4j/cdi/blob/master/api/src/main/java/javax/enterprise/inject/spi/BeanManager.java#L416) and only test propagation if the given context is active and end the test (successfully) otherwise.

Hand in hand with that goes a split of the aforementioned TCK test to check each scope propagation separately. In fact I think we should do that either way. I'm fine sending PR for that, I'll just wait on some opinions here.

EDIT: corrected the link to the test, I meant `CDIContextTest` of course :)

So, WDYT?

manovotn:
Fixes #170 

I split the test and replaced any direct injection with dynamic resolution to avoid trying to inject beans from inactive contexts. Every test now checks if context is active and should it get negative answer, it successfully ends.

As for the change in `mp_checkstyle_rules.xml` - I am fine reverting that if you ask for it, but if you wouldn't mind removing it, I think it's a bit overzealous rule. This is probably the sole project where I am seeing this right curly limitations and checkstyle is giving me hell every time :)

### DIFF
--- a/mp_checkstyle_rules.xml
+++ b/mp_checkstyle_rules.xml
@@ -69,9 +69,6 @@
         <module name="LeftCurly">
             <property name="option" value="EOL"/>
         </module>
-        <module name="RightCurly">
-            <property name="option" value="ALONE"/>
-        </module>
         <module name="EmptyStatement"/>
         <module name="EqualsHashCode"/>
         <module name="DefaultComesLast"/>
--- a/tck/src/main/java/org/eclipse/microprofile/context/tck/cdi/CDIContextTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/context/tck/cdi/CDIContextTest.java
@@ -19,12 +19,19 @@
 package org.eclipse.microprofile.context.tck.cdi;
 
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
 
 import java.lang.reflect.Method;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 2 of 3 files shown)


manovotn:
@kenfinnigan @mkouba just an FYI

PR #175, opened 2019-09-05
Remove circular bean dependency in tests.

manovotn:
This PR has no effect on the tests themselves, it just removes a circular bean dependency from the tests by moving the producers/disposer to a different bean that the one consuming them.

While circular dependencies are supported in CDI (to certain extent), they are not a nice practice, so I thought I'd improve that.


mkouba:
Looks good.

PR #179, opened 2019-09-17
Rewrite TC/ME JTA test to use APIs while still allowing for m…

### LINKED ISSUE #178: TCK dependency on certain EE APIs
In the current state, the TCKs try to avoid dependency on some APIs (CDI and transactions) and use reflection to test that certain functionality works.
A good example is [`ThreadContextTest#clearTransactionContextJTA()`](https://github.com/eclipse/microprofile-context-propagation/blob/master/tck/src/main/java/org/eclipse/microprofile/context/tck/ThreadContextTest.java#L134).

Apparently somewhere in the past we decided to do it that way, probably so that some implementations can implement MP CP without support for certain features. I don't really see a strong argument for that now as we have two impls, both of which use CDI in one way or another and both have transactions existing and either want or don't want to propagate their contexts. Plus we haven't heard of another which would want to go down that road yet.
Plus we now have another integrator (Quarkus) that wants to execute TCKs but has to perform certain CL magic to execute them. This yields problems with the aforementioned test where you get some illegal access exceptions.

I suggest we make that TCK test depend on pure APIs and then use those API methods to determine whether certain functionality is supported. The test can still succeed even if it isn't in place. For instance, with CDI you should be able to do:
```
try {
  CDI<Object> cdi = CDI.current();
  } catch (IllegalStateException e) {
    // no CDI provider, we can end the test here
    return;
  }
cdi.select(...).get().doWhatTheTestNeeds();
```

manovotn:
…issing impls.

Fixes #178 

Rewrites test to eliminate reflections usage and instead uses standard APIs.

### DIFF
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -62,7 +62,6 @@
             <groupId>javax.enterprise</groupId>
             <artifactId>cdi-api</artifactId>
             <version>2.0</version>
-            <scope>provided</scope>
         </dependency>
 
         <dependency>
--- a/tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
@@ -18,9 +18,7 @@
  */
 package org.eclipse.microprofile.context.tck;
 
-import java.lang.annotation.Annotation;
 import java.io.CharConversionException;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.List;
@@ -48,8 +46,12 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
+import javax.enterprise.inject.Instance;
+import javax.enterprise.inject.spi.CDI;
 import javax.naming.InitialContext;
 import javax.naming.NamingException;
+import javax.transaction.Status;
+import javax.transaction.UserTransaction;
 
 import org.eclipse.microprofile.context.tck.contexts.buffer.Buffer;
 import org.eclipse.microprofile.context.tck.contexts.buffer.spi.BufferContextProvider;
@@ -1586,6 +1588,9 @@ public class ManagedExecutorTest extends Arquillian {
      * transaction, but that isn't possible without the TCK having a dependency on a particular
      * type of transactional resource.
      *
+     * The test tries to get hold of {@code UserTransaction} via JNDI and via CDI.
+     * Should neither work, it still doesn't throw exception but instead returns.
+     *
      * @throws Exception indicates test failure
      */
     @Test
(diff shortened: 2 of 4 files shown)


manovotn:
I only found this occurrence in one test within `ThreadContextTest` and once in `ManagedExecutorTest`, correct me if I missed anything...

tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
```diff
@@ -1668,20 +1658,20 @@ public void propagateTransactionContextJTA() throws Exception {
...
-            Assert.assertEquals(getStatus.invoke(tx), 0, // javax.transaction.Status.STATUS_ACTIVE
+            Assert.assertEquals(tx.getStatus(), 0, // javax.transaction.Status.STATUS_ACTIVE
                     "Transaction no longer active after running task.");
 
             // Attempt to propagate this transaction to another thread.
             try {
                 stage2 = stage1.thenApplyAsync(s -> {
                     try {
-                        Assert.assertEquals(getStatus.invoke(tx), 0, // javax.transaction.Status.STATUS_ACTIVE
+                        Assert.assertEquals(tx.getStatus(), 0, // javax.transaction.Status.STATUS_ACTIVE
```
njr-11: Can access Status.STATUS_ACTIVE directly now.

tck/src/main/java/org/eclipse/microprofile/context/tck/ThreadContextTest.java
```diff
@@ -128,79 +129,69 @@ public void builderForThreadContextIsProvided() {
...
-
         ThreadContext threadContext = ThreadContext.builder()
                 .cleared(ThreadContext.TRANSACTION)
                 .unchanged()
                 .propagated(ThreadContext.ALL_REMAINING)
                 .build();
 
         Callable<String> taskWithNewTransaction = threadContext.contextualCallable(() -> {
-            Assert.assertEquals(getStatus.invoke(tx), 6, // javax.transaction.Status.STATUS_NO_TRANSACTION
+            Assert.assertEquals(tx.getStatus(), 6, // javax.transaction.Status.STATUS_NO_TRANSACTION
```
njr-11: Can access Status.STATUS_NO_TRANSACTION directly now.

tck/src/main/java/org/eclipse/microprofile/context/tck/ThreadContextTest.java
```diff
@@ -128,79 +129,69 @@ public void builderForThreadContextIsProvided() {
...
             return "SUCCESS";
         });
 
-        begin.invoke(tx);
+        tx.begin();
         try {
             Assert.assertEquals(taskWithNewTransaction.call(), "SUCCESS");
 
-            Assert.assertEquals(getStatus.invoke(tx), 0, // javax.transaction.Status.STATUS_ACTIVE
+            Assert.assertEquals(tx.getStatus(), 0, // javax.transaction.Status.STATUS_ACTIVE
```
njr-11: Can access Status.STATUS_ACTIVE directly now.

tck/pom.xml
```diff
@@ -62,7 +62,6 @@
             <groupId>javax.enterprise</groupId>
             <artifactId>cdi-api</artifactId>
             <version>2.0</version>
-            <scope>provided</scope>
```
njr-11: Please also remove the unused CDI compile dependency from api/pom.xml

manovotn:
@njr-11 fixes all remarks, thanks for review.

> This isn't necessarily a problem because, as I mentioned earlier, our official run of the TCK is with CDI enabled and that still runs cleanly. I'm more pointing it out because it demonstrates what will happen to an implementation that doesn't provide a CDI implementation or spec. If you haven't already, I'd recommend getting a review from @FroMage as well to confirm this won't cause any problems for RestEasy.

I think it also depends on what exactly is a "CDI disabled" mode? The TCK itself now depends on CDI API and what I tried to do here is that so long as the implementation will have the API available (not the impl), then the tests should be runnable. What you described seems to only be possible in multi CL environment where you deliberately cut out some pieces. Or am I missing something else?

And I did mark Stephane for review already, not planning on merging without his review.

manovotn:
@FroMage whenever you have a minute, please add a review :)

FroMage:
Sorry about the delay. Merged.

PR #180, opened 2019-09-24
generic contextualizer methods to support reactive streams

### LINKED ISSUE #166: Support ReactiveStreams
Although the plugin system of MP-CP and RxJava2 make it possible to have automatic context propagation for RxJava2, it appears some ReactiveStreams implementations don't use RxJava2 or Reactor but built their own implementations, which don't necessarily have a plugin system and so must be manually embarked into context propagation.

We could add those methods to `ThreadContext`:

```java
public <T> Subscriber<T> wrapSubscriber(Subscriber<T> subscriber);
```

At the cost of a dependency to ReactiveStreams, but that's an industry standard and has 4 interfaces.  We could add support for the JDK 9's Flow.Subscriber too, but I don't think we can add a dependency to JDK9.

njr-11:
pull fixes #166

This pull makes an attempt at addressing both of the scenarios for reactive streams from #166 in a generic manner that doesn't add reactive streams or Java 11 as compile dependencies to our spec classes.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/context/ThreadContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/context/ThreadContext.java
@@ -480,6 +480,71 @@ public interface ThreadContext {
     <R> Supplier<R> contextualSupplier(Supplier<R> supplier);
 
     /**
+     * <p>Returns a proxy for one or more interfaces, such that methods run with
+     * context that is captured, per the configuration of this <code>ThreadContext</code>,
+     * from the thread that invokes <code>onMethodsOf</code>.</p>
+     *
+     * <p>When interface methods are invoked on the resulting proxy,
+     * context is first established on the thread that will run the method,
+     * and then the method of the provided instance is invoked.
+     * Finally, the previous context is restored on the thread, and the result of the
+     * method is returned to the invoker.</p>
+     *
+     * <p>For example, in the case of,</p>
+     * <pre>
+     * <code>publisher.subscribe(threadContext.onMethodsOf(subscriber))</code></pre>
+     *
+     * <p>the context of the thread that invokes <code>onMethodsOf</code> is made available to the
+     * subscriber's <code>onSubscribe</code>, <code>onNext</code>, <code>onError</code>, and
+     * <code>onComplete</code> methods whenever they are used.</p>
+     *
+     * @param <T>      an interface that is implemented by the instance.
+     * @param <U>      type of supplied instance.
+     * @param instance instance to contextualize.
+     * @return contextualized proxy for interfaces that are implemented by the supplied instance.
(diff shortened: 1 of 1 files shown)


FroMage:
Thanks. 

What happens implementation-wise? I assume we can only proxy interface methods, so if we invoke `onMethodsOf(Flowable)` (`Flowable` is a class, subtype of RS `Publisher`) we can't really get back a `Flowable` so should it throw?

Why do we have two methods? I fail to see the difference between them.

@cescoffier any opinion?

cescoffier:
It should return the same type as the passed instance, or you will get a _naked_ Reactive Streams Publisher. If you do this, the user will recreate another object from this one, which is not very convenient.

Also, does it work with the `subscribesOn` and `publishOn` operator that switch the thread (if needed)?

njr-11:
> What happens implementation-wise? I assume we can only proxy interface methods, so if we invoke `onMethodsOf(Flowable)` (`Flowable` is a class, subtype of RS `Publisher`) we can't really get back a `Flowable` so should it throw?

It would proxy the interface that Flowable implements: Publisher.  That means the subscribe method (the only one on the Publisher interface) would run with context.
This generic method won't be useful in every case.  It was provided to cover the Subscriber pattern, not Publisher.  What part of a Flowable instance were you hoping to apply context to?  That class has an enormous number of methods.

> Why do we have two methods? I fail to see the difference between them.

It is an attempt to address the concern you raised under https://github.com/eclipse/microprofile-context-propagation/issues/166#issuecomment-506729664 , which I've copied here:

> For `Subscriber` it would be possible, because we capture on creation and restore on any method call, but if we wanted to also provide it for `Publisher` to make it automatic for every subscriber we could not because we don't want to capture and restore but just forward any calls to `subscribe` to our `Subscriber` wrapper.

So we have one method to fit the generic pattern of subscriber where you want all interface methods of the supplied instance to run with context, and a second method that fits the particular pattern that you outline for publisher, where instead subscribed instances are auto-contextualized such that their interface methods run with context.  Personally, I didn't yet see a strong case for this second pattern because it seemed more of a minor convenience, but tried to add it to address the concern.

> It should return the same type as the passed instance, or you will get a _naked_ Reactive Streams Publisher. If you do this, the user will recreate another object from this one, which is not very convenient.
> 
> Also, does it work with the `subscribesOn` and `publishOn` operator that switch the thread (if needed)?

This seems to contradict what was asked for under #166.  I'm fine if plans have changed, in which case we should cancel #166 and get a new issue opened to represent whatever the real requirements are.  It sounds like you are wanting MicroProfile Context Propagation to take an existing instance from an external implementation and mutate it such that context is propagated within it.  We can't go that far, but maybe we can define some sort of mechanism by which other implementations can better plug into MicroProfile Context Propagation to get context propagated when they need it.  In the new issue, can you elaborate more fully on what you are trying to achieve?

PR #184, opened 2019-12-05
Move to Jakarta EE GAVs.

manovotn:
Swap from Java EE to Jakarta EE GAVs.
The only occurrence of these should be within our TCKs.


FroMage:
LGTM, but let's wait for @njr-11 's review.

PR #186, opened 2019-12-19
Revert "Switch version back to 1.1" in order to create a micro version update of 1.0 for Jakarta EE dependencies

njr-11:
Reverts eclipse/microprofile-context-propagation#182

Switching back to 1.0 under this pull so that we can create a 1.0 micro version update with dependencies on Jakarta EE rather than Java EE for the TCK.

Once the above is created, a different pull can be used to revert the revert, which will put us back at version 1.1.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -18,12 +18,12 @@
     <parent>
         <groupId>org.eclipse.microprofile.context-propagation</groupId>
         <artifactId>microprofile-context-propagation-parent</artifactId>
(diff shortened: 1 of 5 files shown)


aguibert:
I would suggest creating a 1.0 branch, and then we can re-direct this PR at the 1.0 branch and not have to worry about doing a temporary version swap

aguibert (approved):

PR #194, opened 2020-07-07
better definition of application context

### LINKED ISSUE #192: Spec needs to clarify what Application Context means
From the conversation with @njr-11 , Application Context means the following:

From @njr-11 
>Application context includes the java:comp, java:module, and so forth JNDI namespace of the application component, as well as its thread context class loader..
Transaction context propagation means that the task or completion stage runs under the same transaction of the thread that submits the task or creates the stage. If the task or completion stage wants to roll back the transaction, it can do so with the usual rollback API, or it can mark the transaction for rollback with setRollbackOnly. The spec does not have anything where we auto-rollback due to a failure, which would limit flexibility of implementation by forcing rollback on exception paths.
Transaction context propagation also covers propagation of the lack of any transaction to a thread. So if the original thread was not running in a transaction to start with, then the task or completion stage will also run without any active transaction on the thread.

@Ladicek states

> That sounds nice, but isn't what the MP ConProp spec says, nor what its TCK tests. Specifically, the spec says that "it can determine the thread context class loader as well as the set of resource references that are available for lookup or resource injection", which is quite vague. The TCK test for the APPLICATION context doesn't test JNDI contexts at all (no wonder -- MicroProfile doesn't know anything about JNDI), and its test for the TCCL propagation doesn't really test anything. I'm aware of one MP ConProp implementation that fully passes the TCK, but doesn't propagate TCCL.

The spec doc needs to be updated to clearly specify the meaning.

njr-11:
fixes #192

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
@@ -20,8 +20,10 @@ package org.eclipse.microprofile.context.tck;
 
 import java.io.CharConversionException;
 import java.lang.reflect.Method;
+import java.util.AbstractMap.SimpleEntry;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
@@ -1567,27 +1569,31 @@ public class ManagedExecutorTest extends Arquillian {
             }).get(MAX_WAIT_NS, TimeUnit.NANOSECONDS);
 
             // Use a ManagedExecutor to apply context to an operation that runs on unmanagedSingleThreadExecutor
-            CompletableFuture<Class<?>> cf = executor.completedFuture(1).thenApplyAsync(i -> {
+            CompletableFuture<Map.Entry<ClassLoader, Class<?>>> cf = executor.completedFuture(1).thenApplyAsync(i -> {
                 try {
                     // load a class from the application
                     ClassLoader loader = Thread.currentThread().getContextClassLoader();
-                    Assert.assertEquals(loader, appClassLoader);
-                    return loader.loadClass("org.eclipse.microprofile.context.tck.contexts.label.Label");
+                    return new SimpleEntry<ClassLoader, Class<?>>(
+                            loader,
+                            loader.loadClass("org.eclipse.microprofile.context.tck.contexts.label.Label"));
                 }
                 catch (ClassNotFoundException x) {
                     throw new CompletionException(x);
                 }
             }, unmanagedSingleThreadExecutor);
 
-            Assert.assertEquals(cf.get(MAX_WAIT_NS, TimeUnit.NANOSECONDS), Label.class,
-                    "Could not load class from application's class loader.");
+            Map.Entry<ClassLoader, Class<?>> result = cf.get(MAX_WAIT_NS, TimeUnit.NANOSECONDS);
+            Assert.assertEquals(result.getKey(), appClassLoader);
+            Assert.assertEquals(result.getValue(), Label.class,
+                    "Did not properly load class from application's class loader.");
 
             // Verify that the class loader of unmanagedSingleThreadExecutor was restored after
             // running the previous task.
-            unmanagedSingleThreadExecutor.submit(() -> {
-                ClassLoader loader = Thread.currentThread().getContextClassLoader();
-                Assert.assertEquals(loader, newClassLoader);
+            ClassLoader loaderForSingleThreadExecutor = unmanagedSingleThreadExecutor.submit(() -> {
+                return Thread.currentThread().getContextClassLoader();
             }).get(MAX_WAIT_NS, TimeUnit.NANOSECONDS);
+
+            Assert.assertEquals(loaderForSingleThreadExecutor, newClassLoader);
         }
         finally {
             executor.shutdownNow();
--- a/tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
@@ -1557,8 +1557,17 @@ public class ManagedExecutorTest extends Arquillian {
 
         ManagedExecutor executor = builder.build();
 
+        ExecutorService unmanagedSingleThreadExecutor = Executors.newFixedThreadPool(1);
         try {
-            CompletableFuture<Class<?>> cf = executor.supplyAsync(() -> {
+            // Set the class loader of the single thread upon which unmanagedSingleThreadExecutor runs tasks
+            ClassLoader newClassLoader = unmanagedSingleThreadExecutor.submit(() -> {
+                ClassLoader loader = new ClassLoader() {};
+                Thread.currentThread().setContextClassLoader(loader);
+                return loader;
+            }).get(MAX_WAIT_NS, TimeUnit.NANOSECONDS);
+
+            // Use a ManagedExecutor to apply context to an operation that runs on unmanagedSingleThreadExecutor
+            CompletableFuture<Class<?>> cf = executor.completedFuture(1).thenApplyAsync(i -> {
                 try {
                     // load a class from the application
                     ClassLoader loader = Thread.currentThread().getContextClassLoader();
@@ -1568,13 +1577,21 @@ public class ManagedExecutorTest extends Arquillian {
                 catch (ClassNotFoundException x) {
                     throw new CompletionException(x);
                 }
-            });
+            }, unmanagedSingleThreadExecutor);
 
             Assert.assertEquals(cf.get(MAX_WAIT_NS, TimeUnit.NANOSECONDS), Label.class,
                     "Could not load class from application's class loader.");
+
+            // Verify that the class loader of unmanagedSingleThreadExecutor was restored after
+            // running the previous task.
+            unmanagedSingleThreadExecutor.submit(() -> {
+                ClassLoader loader = Thread.currentThread().getContextClassLoader();
+                Assert.assertEquals(loader, newClassLoader);
+            }).get(MAX_WAIT_NS, TimeUnit.NANOSECONDS);
         }
         finally {
             executor.shutdownNow();
+            unmanagedSingleThreadExecutor.shutdownNow();
         }
     }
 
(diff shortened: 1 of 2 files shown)


manovotn:
Looking at TCKs, we only have [this test](https://github.com/eclipse/microprofile-context-propagation/blob/master/tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java#L1550). We should revisit it too because:
* the test is currently optional only which I am not sure is needed/intentional
* the test only verifies that there is *some* TCCL present, not that it is the same one; e.g. I am not sure the test is actually testing anything (or am I missing something?)

njr-11:
> Looking at TCKs, we only have [this test](https://github.com/eclipse/microprofile-context-propagation/blob/master/tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java#L1550). We should revisit it too because:
> 
>     * the test is currently optional only which I am not sure is needed/intentional
> 
>     * the test only verifies that there is _some_ TCCL present, not that it is the same one; e.g. I am not sure the test is actually testing anything (or am I missing something?)

I'd be happy to switch the test to require application context.  Beyond that, testing for a JNDI namespace would require that we add a Jakarta/Java EE resource reference to the TCK application, which currently doesn't depend on Jakarta/Java EE.  Is that a concern here?  As for the thread context class loader - it is at least being used to load an application class. But I could see some implementations getting around that by always having all classes available.

FroMage:
Huh, I never understood what application-scoped meant, which is why we don't have an application-scoped context provider in Quarkus.
I guess there's value in providing propagation for TCCL (we've had users ask for that), but I wonder what will happen if we make this the default (could be some stuff breaks, but I hope not).
We don't support JNDI though, so we should not request the application context to propagate JNDI even if the platform doesn't support JNDI.

TBH I find these predefined contexts just don't match with the real world outside of EE, because in Quarkus we have some request objects in the CDI context, others in the RESTEasy context and yet others in the Servlet context. Our security is using CDI, so we have no Security context (it's already in the CDI one), and the TCCL is actually a constant since we can only have a single application, so capturing/restoring it doesn't make a lot of sense in most situations where the framework manages the threads, because we can simply make sure to always set it for those threads, without MP-CP being involved. This is less true if the threads are not managed by the framework.

So, I'm fine with asking that the `Application` context propagates the TCCL (we do have to change the test, it currently only checks that the thread can load the same classes as the test itself, and I suspect it's only passing in Quarkus because our thread pool has the TCCL set, not by MP-CP).
But I'd don't want to specify that the `Application` must propagate JNDI if it's not available.

manovotn:
> TBH I find these predefined contexts just don't match with the real world outside of EE, because in Quarkus we have some request objects in the CDI context, others in the RESTEasy context and yet others in the Servlet context. Our security is using CDI, so we have no Security context (it's already in the CDI one), and the TCCL is actually a constant since we can only have a single application, so capturing/restoring it doesn't make a lot of sense in most situations where the framework manages the threads, because we can simply make sure to always set it for those threads, without MP-CP being involved.

This nicely sums up that what context propagation means can easily differ between platforms using different implementations. And the harder we try to pre-define it and set defaults, the less realistic it can become for wider use (in between several platforms).

> So, I'm fine with asking that the Application context propagates the TCCL (we do have to change the test, it currently only checks that the thread can load the same classes as the test itself, and I suspect it's only passing in Quarkus because our thread pool has the TCCL set, not by MP-CP).
But I'd don't want to specify that the Application must propagate JNDI if it's not available.

+1

njr-11:
@manovotn I added a commit to update the test case per your review comments. Thanks for spotting that!

tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
```diff
@@ -1548,23 +1548,20 @@ public void newIncompleteFutureDependentStagesRunWithContext() throws ExecutionE
...
-            return; // Skip test if Application context is not supported.
-        }
+        ManagedExecutor executor = builder.build();
 
         try {
             CompletableFuture<Class<?>> cf = executor.supplyAsync(() -> {
                 try {
                     // load a class from the application
                     ClassLoader loader = Thread.currentThread().getContextClassLoader();
+                    Assert.assertEquals(loader, appClassLoader);
```
Ladicek, line 1564: This still doesn't test anything, does it? This just tests that the new thread has the same TCCL as the original thread, which can easily happen without any propagation happening.

FroMage: Yeah. I was surprised in my tests to realise that all my new threads had a TCCL set to the JDK CL. Not sure who sets them.

FroMage:
So, the question came up of whether it's safe/desirable to set the TCCL to `null` in `ThreadContextProvider.clearedContext`? WDYT?
ATM the JDK sets the TCCL for every new thread, so this could be dangerous, but I'm not sure how we can obtain that default TCCL if that's what we want for the "cleared" semantics.

manovotn:
Looking at Stephane's PR for SR, I was thinking - what should happen when you say you want application context `cleared`?
Namely I am looking at TCCL.

First thing that comes to mind is to just set it to `null`. However, I think that is very rarely useful and can even be harmful in some cases. Shouldn't we specify that we just "ignore" it in that case?

FroMage:
Javadoc says this about the initial TCCL:

> Returns the context ClassLoader for this thread. The context ClassLoader is provided by the creator of the thread for use by code running in this thread when loading classes and resources. If not set, the default is the ClassLoader context of the parent thread. The context ClassLoader of the primordial thread is typically set to the class loader used to load the application.

So it's not necessarily trivial to obtain the right initial TCCL to restore it to, but `null` ain't it.

njr-11:
I checked on what the OpenLiberty implementation does when clearing application context, and it sets the system classloader, `ClassLoader.getSystemClassLoader()`.  This same code is also used in our implementation of Java EE Concurrency, which has been around for years, and we have yet to see anyone raise a complaint over it.  My guess is that this choice was match to match the behavior of setting a null thread context class loader.  As specified by java.lang.Thread, null indicates `the system class loader (or, failing that, the bootstrap class loader)`.

Given that managed executors allow for the use of pooled threads, some of which might be created lazily, possibly with different class loaders on the parent thread at the time, and we want deterministic usage in all cases, there isn't really a proper default TCCL to clear to.  Ignoring rather than clearing the thread context class loader seems like a bug to me, because if a user specified they want it cleared, it could mean they don't want whatever code they are submitting/scheduling/invoking to be able to load their application classes.

What are the cases where you see a cleared thread context class loader being harmful?  And why wouldn't the user just specify to `propagate` or `ignore` in those cases?

manovotn:
I have seen some cases (bugs) on WFLY where they were relying on TCCL never being `null` and that failed for instance in case of async events with Weld (where we provide executor with nulled TCCL intentionally). So you might be right that it is indeed correct because I found the same javadoc notion of `null` being a valid value meaning basically "system CL".

tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
njr-11: @Ladicek @FroMage  I added a commit that enhances the test case to ensure the class loader is propagated (and restored afterward).   This switches the test case to execute the completion stage on an unmanaged thread, but with context, where the unmanaged thread has a predetermined class loader that is not the same.  If the context is successfully propagated, then the requesting code's class loader will appear on that thread instead of the predetermined one, and, once it finishes running the completion stage with context, the class loader of that thread will go back to the predetermined one, which the test case also checks for.  Hopefully this will be sufficient.

tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
```diff
@@ -1549,35 +1549,49 @@ public void newIncompleteFutureDependentStagesRunWithContext() throws ExecutionE
...
+            }, unmanagedSingleThreadExecutor);
 
             Assert.assertEquals(cf.get(MAX_WAIT_NS, TimeUnit.NANOSECONDS), Label.class,
                     "Could not load class from application's class loader.");
+
+            // Verify that the class loader of unmanagedSingleThreadExecutor was restored after
+            // running the previous task.
+            unmanagedSingleThreadExecutor.submit(() -> {
+                ClassLoader loader = Thread.currentThread().getContextClassLoader();
+                Assert.assertEquals(loader, newClassLoader);
```
Ladicek, line 1589: This assert needs to be moved to the "main" test thread. Exceptions that are thrown on other threads don't contribute to test status.

tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
Ladicek: The approach looks good, just added one small but important comment.

tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
njr-11: I'm surprised to hear of a behavior like that so I tried it out by temporarily specify a value for the assert to intentionally make it fail.  The error reported in the test results for me just fine.  Nonetheless, I cannot rule out the possibility that it differs based on the environment in which you are running, so the safe thing to do is to move the asserts to the main test thread.  I have done that under commit https://github.com/eclipse/microprofile-context-propagation/pull/194/commits/f95439d40d62f35de3c9cd03fe6f1e2aa2778f71 - please review again.

Ladicek: Weird, I've seen tests passing when a failing assert happened on an extra thread. But now I realize that was with JUnit, and this is TestNG, right? Perhaps TestNG takes care of that -- which would certainly be nice, I didn't know about that.

Ladicek (approved):
This looks good to me, thanks for your patience :-)

tck/src/main/java/org/eclipse/microprofile/context/tck/ManagedExecutorTest.java
njr-11: Yeah, hopefully TestNG covers it better.  The other possibility I can think of would be if the failing assert on the separate thread happened _after_ the main test method completed.  That would never happen in our particular case because the main thread always waits for the completion stage that is running async to finish before exiting the test method.

PR #195, opened 2020-07-13
readme section for linking to vendor defaults

njr-11:
The pull adds a section to the readme mentioning the lack of defaults for context propagation in the spec and how to cope with this.  A comment line indicates where we can add links to vendor documentation once it becomes available. That will be done in a separate pull.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -177,17 +177,15 @@ by implementations that are also Jakarta/Java EE application servers.
 
 // TODO: insert links to implementation defaults here, once available
 
-Applications can achieve portability across implementations with respect to propagated
-context types either by explicitly specifying the context propagation on the builder
-or by explicitly configuring the desired defaults via MicroProfile Config properties.
+You can define your own defaults for context propagation via MicroProfile Config properties.
 For example,
 
 [source,java]
 ----
-mp.context.ManagedExecutor.propagated=Application,CDI,Security
-mp.context.ManagedExecutor.cleared=Remaining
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -163,6 +163,33 @@ to define and manage single instances across an application:
...
+
+Applications can achieve portability across implementations with respect to propagated
+context types either by explicitly specifying the context propagation on the builder
+or by explicitly configuring the desired defaults via MicroProfile Config properties.
+For example,
+
+[source,java]
+----
+mp.context.ManagedExecutor.propagated=Application,CDI,Security
+mp.context.ManagedExecutor.cleared=Remaining
```
FroMage, line 188: So, in general this sort of configuration can be problematic for Quarkus users because RESTEasy and Servlet have their own context that users may be surprised to learn will not be in the CDI context.
Perhaps explicitely specifying cleared contexts is less dangerous?

njr-11: Explicitly specifying cleared contexts would be less disruptive in terms of continuing to use the implementation you are already on without trouble, but this seems like heading in the opposite direction to achieving portability because the RESTEasy context that the application wanted to assume was available has no guarantee of being available on another implementation.  With it showing in the propagated list, the application is at least stating its full requirements for an implementation that would need to be met by an implementation that is compatible.

That said, I can still switch to what you suggested.  I'll just need to remove the language about it being a way to achieve portability.  I'll add a second commit with that.

PR #198, opened 2020-09-25
work around mpconfig dropping support for empty array

### LINKED ISSUE #160: MP Config - introduce constant for empty string
Currently we allow users to set their defaults via MP config in the following way - `ThreadContext/propagated=`.
This is interpreted as "no contexts are to be propagated".

Now, looking at the changes proposed in https://github.com/eclipse/microprofile-config/issues/397 and the respective PR https://github.com/eclipse/microprofile-config/pull/407, we are going to have to change this.
With the above, empty value means you won't be able to see the key is there at all.

Hence I would suggest some constant, much like we have the `REMAINING` one, but this one would have to be something like `none`?

What do you guys think?

njr-11:
contains most of the fix for #160

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/context/ThreadContext.java
+++ b/api/src/main/java/org/eclipse/microprofile/context/ThreadContext.java
@@ -303,11 +303,19 @@ public interface ThreadContext {
     static final String CDI = "CDI";
 
     /**
-     * An empty array of thread context.
-     * This is provided as a convenience for code that wishes to be more explicit.
+     * <p>An empty array of thread context.</p>
+     *
+     * <p>This is provided as a convenience for code that wishes to be more explicit.
      * For example, you can specify <code>builder.propagated(ThreadContext.NONE)</code>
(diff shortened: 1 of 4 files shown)


Emily-Jiang (approved):
LGTM!

manovotn (approved):
Sorry for the delay, was on PTO.
This is fine with me.

FroMage (approved):
LGTM, but should we forbid `ThreadContextProvider` with the name `None` to be extra sure we'll never run into issues? I guess MP config will always filter them out, but still…

njr-11:
> LGTM, but should we forbid `ThreadContextProvider` with the name `None` to be extra sure we'll never run into issues? I guess MP config will always filter them out, but still…

It seems extremely unlikely that anyone would name their ThreadContextProvider "None", but it's easy enough to explicitly forbid that so we might as well.  I'll add it in a separate pull under the same issue given that there are already 3 approvals on this one.

PR #211, opened 2021-08-02
Issue #210 update TCK and doc for Jakarta EE 9 packages

njr-11:
Updates the TCK and a couple of code examples in the documentation to use jakarta packages.
This is hopefully all we need to change with respect to code for the Jakarta EE 9 alignment effort.
See #210 for more detail.

### DIFF
--- a/mp_checkstyle_rules.xml
+++ b/mp_checkstyle_rules.xml
@@ -51,7 +51,7 @@
         </module>
         <module name="AvoidStarImport">
             <property name="excludes"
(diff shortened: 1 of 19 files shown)


FroMage (approved):
LGTM.

njr-11:
I tested out the jakarta-based TCK against Open Liberty, where it passes at 86%.  It should be noted that some tests are expected to fail because they require MP Config, of which a release that supports jakarta packages does not yet exist.  So this looks good, and I'll also look into creating an RC build that others can experiment with as well.

PR #218, opened 2022-02-03
Use MP Parent POM

radcortez:


njr-11 (approved):
This looks great - thanks so much for getting all of these updates in to use the parent pom!
The only question I have is about the version, whether to use 1.4 or 2.0.  At this point, it's just a guess, so your switch to 2.0 should be fine and could always be switched back to 1.4 if the next release ends up being a minor one.  2.0 seems a good guess for eventual alignment with Jakarta EE 10.

radcortez:
> The only question I have is about the version, whether to use 1.4 or 2.0. At this point, it's just a guess, so your switch to 2.0 should be fine and could always be switched back to 1.4 if the next release ends up being a minor one. 2.0 seems a good guess for eventual alignment with Jakarta EE 10.

I actually think we should do a 2.0 release right now for consistency reasons. All other MP projects bumped their major version when they switched to Jakarta, but here we only bumped the minor.

PR #222, opened 2026-01-06
Added `module-info.java`

### LINKED ISSUE #221: Assign a JPMS module name
The spec API does not have a defined JPMS module name.

I would recommend defining an automatic module name of `org.eclipse.microprofile.context`, since that is the name of the top-level package of this project.

FroMage:
Also bumped to latest microprofile-parent otherwise it did not compile.

Fixes #221

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -28,4 +28,19 @@
     <name>MicroProfile Context Propagation</name>
     <description>MicroProfile Context Propagation :: API</description>
 
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-checkstyle-plugin</artifactId>
+                    <configuration>
+                        <!-- https://github.com/checkstyle/checkstyle/issues/8240 -->
+                        <excludes>**/module-info.java</excludes>
+                    </configuration>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+    </build>
+
 </project>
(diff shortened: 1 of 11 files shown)


FroMage:
Could you try this @dmlloyd ?

FroMage:
@radcortez @njr-11 What are our options for making a release now that this spec has been deprecated?

dmlloyd:
Looks good 👍🏻

FroMage:
Not "looks good"… works? 😅

dmlloyd:
As far as I can tell. My capacity to test it thoroughly is limited for now.

dmlloyd (approved):

radcortez (approved):

radcortez:
> @radcortez @njr-11 What are our options for making a release now that this spec has been deprecated?

We can release as usual; being deprecated is just information to the community that we are not actively developing or maintaining it. 

It would be easier if we could release it as a service release. Let's see.

njr-11:
In addition to being deprecated, it is listed as discontinued on its git project page, https://github.com/microprofile/microprofile-context-propagation
I wouldn't expect any further releases of it unless somehow a security vulnerability was discovered in it.  And right, that would be a service release only.

FroMage:
@njr-11 So, you would be opposed to a release with a module declaration?

This feels more like a metadata fix for newer Java releases than a new feature to me. But it's not a security fix, for sure.

njr-11:
I don't see a reason to oppose it.  But before someone goes through all the work of creating a maintenance release, it might be a good idea to confirm if there are any users who are needing it versus someone noticing that a module declaration was absent and reporting it.

dmlloyd:
We need it for Quarkus. We're working on modular deployment modes, and our modules need stable names at a minimum.

FroMage:
@njr-11 yeah, we're not doing this because we're looking for things to do 😅 We actually need something like this.

